PR #1, opened 2020-06-01
Document proposed design

72636c:
I’ve already reached semantic satiation with the name.


72636c:
I was thinking that Wingman could deploy its own instance with playground credentials, but we’d have to double check that everything there can be made fully public.

PR #4, opened 2020-06-02
Add createSeekGraphMiddleware

72636c:
This is the first cab off the rank as it doesn’t depend on external infrastructure. It’s slightly restructured from our first cut at this functionality:

- Auth exchange is handled by a central `getPartnerToken` function instead of being bolted on as a separate middleware on the same path
- A `graphql-tools` executor replaces Apollo links
- A middleware is returned, which seems nicer than `applyMiddleware`

### DIFF
--- a/be/src/seekGraph/middleware.test.ts
+++ b/be/src/seekGraph/middleware.test.ts
@@ -1,0 +1,97 @@
+import { Server } from 'http';
+
+import { AuthenticationError } from 'apollo-server-koa';
+import Koa from 'koa';
+import nock from 'nock';
+import request from 'supertest';
+
+import { INTROSPECTION_RESPONSE } from '../testing/graphql';
+
+import { SEEK_API_BASE_URL, SEEK_API_PATH } from './constants';
+import { createSeekGraphMiddleware } from './middleware';
+import { SeekGraphContext } from './types';
+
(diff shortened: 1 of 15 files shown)


be/src/seekGraph/middleware.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      .send({ query: '{ _query }' })
+      .expect(200);
+
+    expect(response.body).toEqual({
+      data: {
+        _query: null,
+      },
+      errors: [
+        expect.objectContaining({
+          /* eslint-disable-next-line @typescript-eslint/no-unsafe-assignment */
```
72636c, line 86: This is lovely

etaoins: This also works if you `as unknown` it before assigning it.

Another option is doing something like this in `eslintrc` while Jest fixes their types up:
```yaml
overrides:
  - files: '**/*.test.ts'
    rules:
      # Allow `any` in our tests
      '@typescript-eslint/no-unsafe-assignment':
        - off
      '@typescript-eslint/no-unsafe-member-access':
        - off
      '@typescript-eslint/no-unsafe-call':
        - off
      '@typescript-eslint/no-unsafe-return':
        - off
```

PR #6, opened 2020-06-03
Add createSeekAttachmentMiddleware

72636c:
This configures a proxy handler for attachment downloads. The major difference compared to the original implementation is that this has been rebased on `node-fetch` and passes a readable stream through to Koa.

I’ve tried to reuse the same `GetPartnerToken` definition here but I think the cracks are starting to show. Once the middlewares have settled a little, we can revisit its signature.

### DIFF
--- a/be/src/getPartnerToken.ts
+++ b/be/src/getPartnerToken.ts
@@ -1,0 +1,11 @@
+/**
+ * Function that verifies that the incoming request is authorised to act on
+ * behalf of a partner and provides a partner token in response.
+ */
+export type GetPartnerToken = (
+  request: GetPartnerTokenRequest,
+) => Promise<string>;
+
+export interface GetPartnerTokenRequest {
+  authorization?: string;
+}
--- a/be/src/seekAttachment/headers.ts
+++ b/be/src/seekAttachment/headers.ts
@@ -1,0 +1,22 @@
+import { Headers } from 'node-fetch';
+
+const HEADERS_TO_FORWARD = {
+  'Content-Disposition': 'attachment',
+  'Content-Length': undefined,
+  'Content-Type': 'application/octet-stream',
+  'Last-Modified': undefined,
+} as const;
+
+export const filterHeaders = (headers: Headers) =>
+  Object.entries(HEADERS_TO_FORWARD).reduce<Record<string, string>>(
+    (subset, [name, defaultValue]) => {
+      const value = headers.get(name) ?? defaultValue;
+
+      if (typeof value !== 'undefined') {
+        subset[name] = value;
+      }
+
+      return subset;
+    },
+    {},
+  );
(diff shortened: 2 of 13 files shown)


be/src/getPartnerToken.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Function that verifies that the incoming request is authorised to act on
+ * behalf of a partner and provides a partner token in response.
+ */
+export type GetPartnerToken = (
+  request: GetPartnerTokenRequest,
+) => Promise<string>;
```
72636c, line 7: Change from `Promise<string | undefined>` is breaking, but we don’t have a release yet 🙂

be/src/seekAttachment/headers.ts
```diff
@@ -0,0 +1,22 @@
+import { Headers } from 'node-fetch';
+
+const HEADERS_TO_FORWARD = {
```
etaoins, line 3: I like this pattern :+1:

be/src/seekAttachment/middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+  const { url } = ctx.query as Record<string, unknown>;
+
+  if (typeof url !== 'string') {
+    return ctx.throw(400, "Query parameter 'url' must be a string");
+  }
+
+  const { origin } = new URL(url);
+
+  if (origin !== SEEK_API_BASE_URL) {
+    return ctx.throw(400, "Query parameter 'url' must point to the SEEK API");
```
etaoins, line 38: :+1:

be/src/seekAttachment/middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+  } catch (err) {
+    // This is a bit of a hack. Consider either exposing the full Koa context to
+    // `getPartnerToken`, or standardising error behaviour so that we don’t
+    // handle unexpected errors and expose internal workings to the client.
+    return ctx.throw(401, err);
+  }
+};
+
+const parseUrlParameter = (ctx: Context) => {
+  const { url } = ctx.query as Record<string, unknown>;
```
etaoins, line 29: Hmm, I'm guessing this can be used to make a `GET` request to any URL on `graphql.seek.com` with a partner token? Is that existing behaviour?

72636c: Yeah it is unfortunately. Do you have a better idea than very strict path matching on known attachment routes?

etaoins: I was thinking that partners would specify something like the `CandidateProfile` ID and attachment index and the server would request that. I can see why they wouldn't though - that would be annoying.

This seems like a good motivator for the hypothetical `download:candidate-profile-attachments` browser token scope.

72636c: There’s more to flesh out there, like validating that the user has access to that `CandidateProfile`. I’ll go with a pattern match for now and include some comments.

PR #12, opened 2020-06-03
Serve attachments from example backend

72636c:
And now the download links do something, albeit poorly!

### DIFF
--- a/be/server/src/config.ts
+++ b/be/server/src/config.ts
@@ -1,0 +1,8 @@
+export const PORT = '9090';
+
+export const FE_ORIGIN = 'http://dev.seekunifiedbeta.com:8080';
+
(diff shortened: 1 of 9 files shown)


be/server/src/config.ts
```diff
@@ -0,0 +1,12 @@
+export const PORT = '9090';
+
+export const FE_ORIGIN = 'http://dev.seekunifiedbeta.com:8080';
+
+export const USER_AGENT = 'wingman/local';
+
+// Take care when handling your token; it’s easy for environment variables to be
+// accessed by other processes, accidentally logged, or persisted to shell
+// history. We recommend wiring this up to a secret store like AWS Secrets
```
etaoins, line 9: :+1:

PR #14, opened 2020-06-03
Retrieve version via SEEK proxy endpoint

72636c:
My CORS woes magically dissipated.

This uses raw Fetch API at the moment; I’m going to play with Apollo Client 3 to see if it’s viable.

---

Screen Shot 2020-06-03 at 20 15 27


72636c:
Apparently they’ve gotten rid of that weird `apollo-boost` name, so I’m already sold

PR #16, opened 2020-06-03
Add createPartnerWebhookMiddleware stub

72636c:
This only formally covers input validation and signature verification, though storage could be implemented through the callback.

It also wires up the example frontend and backend with a new feature: the frontend can POST a fake webhook. If we can’t figure out a nice way to get the webhook interaction working in a self-contained manner, we may just support some form of seeding. This could be triggered at startup and/or by user interaction.

---

Screen Shot 2020-06-03 at 23 56 41

### DIFF
--- a/be/example/src/config.ts
+++ b/be/example/src/config.ts
@@ -4,9 +4,19 @@ export const FE_ORIGIN = 'http://dev.seekunifiedbeta.com:8080';
 
 export const USER_AGENT = 'wingman/local';
 
-// Take care when handling your token; it’s easy for environment variables to be
-// accessed by other processes, accidentally logged, or persisted to shell
-// history. We recommend wiring this up to a secret store like AWS Secrets
-// Manager.
+/**
+ * Take care when handling your token; it’s easy for environment variables to be
+ * accessed by other processes, accidentally logged, or persisted to shell
+ * history. We recommend wiring this up to a secret store like AWS Secrets
+ * Manager.
+ */
 export const getPartnerToken = () =>
   Promise.resolve(process.env.DANGEROUS_PLAYGROUND_TOKEN ?? '');
+
+/**
+ * This ignores signature verification for all incoming webhooks! Don’t do this
+ * in production. Store the signing secret against each subscription ID in a
+ * secure place that you can look up from your software, such as an encrypted
+ * database or secret store.
+ */
+export const getSigningSecret = () => Promise.resolve(null);
--- a/be/src/partnerWebhook/signature.ts
+++ b/be/src/partnerWebhook/signature.ts
@@ -1,0 +1,27 @@
+/* eslint-disable new-cap */
+
+import crypto from 'crypto';
+
+interface ValidateWebhookBodyOptions {
+  body: string;
+  secret: string;
+  signature: string;
+}
+
+export const validateWebhookSignature = ({
+  body,
+  secret,
+  signature,
+}: ValidateWebhookBodyOptions) => {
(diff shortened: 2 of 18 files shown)


be/example/src/config.ts
```diff
@@ -4,9 +4,19 @@ export const FE_ORIGIN = 'http://dev.seekunifiedbeta.com:8080';
...
+ * Take care when handling your token; it’s easy for environment variables to be
+ * accessed by other processes, accidentally logged, or persisted to shell
+ * history. We recommend wiring this up to a secret store like AWS Secrets
+ * Manager.
+ */
 export const getPartnerToken = () =>
   Promise.resolve(process.env.DANGEROUS_PLAYGROUND_TOKEN ?? '');
+
+/**
+ * This ignores signature verification for all incoming webhooks! Don’t do this
```
etaoins, line 17: @zsims has really gotten to you. Should we strengthen the language on the Developer Site to encourage signature verification?

fe/example/src/api/partnerWebhook.ts
```diff
@@ -0,0 +1,26 @@
+import { BE_BASE_URL } from './constants';
+
+const webhookUrl = (endpoint: string) => `${BE_BASE_URL}/webhook/${endpoint}`;
+
+/**
+ * You wouldn't have this in a real production system, but it's an easy way to
+ * seed data for demonstrative purposes without having to deploy a database,
+ * public webhook endpoint, etc.
+ */
+export const postPartnerWebhook = async (data: unknown, endpoint: string) => {
```
etaoins, line 10: :tada:

This is really good. We definitely need more ways to demonstrate webhooks before going fully prod.

I keep toying around with the idea of adding a mutation that would send a `CandidateApplicationCreated` for a given application ID to once-off URL. I suspect that would be a lot of work if we wanted to share code paths with real webhooks, though.

be/src/partnerWebhook/signature.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable new-cap */
```
etaoins, line 1: I wonder if we should start linking out to GitHub from the Developer Site once the churn dies down? Similar to GraphQL we could have the existing TypeScript example codeblock open this file in a new tab.

fe/example/src/api/partnerWebhook.ts
etaoins: Hmm, actually would this work:

1. Create a new e.g. `sendTestCandidateApplicationCreatedEvent` that takes a candidate OID and URL. Presumably there would be other `sendTest`... mutations later. This would duplicate the code to create the event body. It would probably need to go through our candidate application service's public test instance to validate and lookup the IDs.

2. Add a new `submitTestEvent` endpoint in our webhook service that takes a URL instead of a hirer ID. It would run all the normal security validation on the URL as if you were creating a new subscription.

3. Have our webhook service invoke our HTTP caller Lambda synchronously (skip persisting anything to the database). We could set a concurrency limit of 1 to avoid people spamming/DOSing through out caller (even though this would be Playground authenticated)

4. Return a pseudo-delivery attempt back to GraphQL

?

be/example/src/config.ts
72636c: Signed webhooks vs unsigned webhooks is the new webhooks vs polling

fe/example/src/api/partnerWebhook.ts
72636c: That sounds feasible. Resolving the OID and constructing the event body will be the sketchy part.

PR #26, opened 2020-06-04
Fix version command, again

72636c:
Because `yarn version` prints the literal Yarn version 🤡


72636c:
Actually I think I've misdiagnosed what went wrong; closing until I have a better idea

PR #31, opened 2020-06-05
Add createBrowserTokenMiddleware

72636c:
This is better practice than using the SeekGraph proxy as the access is scoped to a hirer, as it would be in an actual third-party system. The interface is a bit rough but hopefully the core functionality is there; I'll follow this up with an actual implementation in the example frontend.

### DIFF
--- a/be/src/browserToken/middleware.ts
+++ b/be/src/browserToken/middleware.ts
@@ -1,0 +1,93 @@
+import { Middleware } from 'koa';
+import bodyParser from 'koa-bodyparser';
+import compose from 'koa-compose';
+import LRU from 'lru-cache';
+import fetch from 'node-fetch';
+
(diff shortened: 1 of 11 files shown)


be/src/browserToken/middleware.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const expiresInMs = responseResult.value.expires_in * 1000;
+
+    const freshToken: BrowserTokenResponse = {
+      authorization: `Bearer ${responseResult.value.access_token}`,
+      expiry: new Date(Date.now() + expiresInMs).toISOString(),
+    };
+
+    // Write off the token at its half life. This is a bit mean.
+    tokenCache.set(hirerId, freshToken, expiresInMs / 2);
```
etaoins, line 80: :+1:

PR #32, opened 2020-06-05
feat: Add Storybook with example

aaronvanston:
# Overview
- Adds storybook into wingman front-end with example component set-up and structure
- Adds automatic deploy of storybook, unsure how this will play with changeset


.changeset/big-poems-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': major
```
72636c, line 2: Could we `minor` this? I'd like to hold on to the v0 for now 😅

.changeset/big-poems-attend.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+feat: add storybook with Example component structure
```
72636c, line 5: We don't need the semantic prefix here 🙂

fe/package.json
```diff
@@ -12,7 +13,9 @@
     "react-dom": "16.13.1",
     "sku": "10.3.5"
   },
-  "files": ["lib"],
+  "files": [
+    "lib"
+  ],
   "license": "MIT",
   "main": "lib/components/index.ts",
```
72636c, line 20: Ah, we'll need to repoint this to `lib/index.ts`?

.github/workflows/release.yml
```diff
@@ -35,3 +35,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.SEEK_OSS_CI_NPM_TOKEN }}
+
+      - name: Publish Storybook
+        run: yarn fe storybook:deploy -- --ci --host-token-env-variable=GITHUB_TOKEN
```
72636c, line 40: I assume this will deploy to `/docs` by default?

aaronvanston: To confirm, it builds into `/dist-storybook` which is then deployed to a `gh-pages` branch.

PR #41, opened 2020-06-05
Switch backend example onto skuba

72636c:
`skuba`, much like `sku`, doesn't seem to play perfectly in a monorepo. This is probably a good way for us to dogfood the use case.


72636c:
Requires https://github.com/seek-oss/skuba-dive/pull/4

be/example/src/app.ts
```diff
@@ -1,9 +1,32 @@
...
+  return {
+    callback: (): RequestListener => (req, res) =>
+      callbackPromise.then((callback) => callback(req, res)),
+
+    listen: (port?: number, listeningListener?: () => void) =>
+      appPromise.then((app) => app.listen(port, listeningListener)),
+
+    port: PORT,
+  };
+};
```
72636c, line 28: This little dance is pretty gross and annoying. I'm wondering if we should bundle it in Koala?

72636c: Just realised I was overcomplicating this for no reason. I'll still package it up as a Koala proposal and see where it goes.

PR #45, opened 2020-06-07
Apply fixes to browser token middleware

72636c:
The middleware now points to the browser token playground URL, and accepts a `callback` for debugging/logging.

### DIFF
--- a/be/src/constants.ts
+++ b/be/src/constants.ts
@@ -3,5 +3,7 @@ export const SEEK_API_BASE_URL = 'https://graphql.seek.com';
 export const SEEK_API_PATH = '/graphql';
 export const SEEK_API_URL = `${SEEK_API_BASE_URL}${SEEK_API_PATH}`;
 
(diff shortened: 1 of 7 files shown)


be/src/constants.ts
```diff
@@ -3,5 +3,7 @@ export const SEEK_API_BASE_URL = 'https://graphql.seek.com';
 export const SEEK_API_PATH = '/graphql';
 export const SEEK_API_URL = `${SEEK_API_BASE_URL}${SEEK_API_PATH}`;
 
+export const SEEK_BROWSER_TOKEN_PLAYGROUND_BASE_URL =
+  'https://indie-jessiej-public-test.public.indirect-apply.prod.outfra.xyz';
```
72636c, line 7: 😞

etaoins: Ahhh yes I forgot about that. We need to fix that.

PR #46, opened 2020-06-07
Use browser tokens for example downloads

72636c:
- Builds on #45
- Is currently blocked by CORS

### DIFF
--- a/fe/lib/hooks/BrowserToken/types.ts
+++ b/fe/lib/hooks/BrowserToken/types.ts
@@ -1,0 +1,6 @@
+export interface BrowserToken {
+  authorization: string;
+  expiry: string;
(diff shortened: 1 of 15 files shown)


fe/lib/hooks/BrowserToken/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface BrowserToken {
+  authorization: string;
+  expiry: string;
+}
```
72636c, line 4: This is the structure that the backend middleware currently uses, but we should probably revisit it. It might be simpler to:

- Pass the original SEEK API structure through the backend
- Simplify this frontend interface so the hook handles expiry and doesn't expose it

PR #49, opened 2020-06-08
Handle browser tokens better

72636c:
- https://github.com/seek-oss/wingman/pull/46#discussion_r436318560
- We weren't caching by scope

### DIFF
--- a/be/src/browserToken/middleware.ts
+++ b/be/src/browserToken/middleware.ts
@@ -13,10 +13,22 @@ import {
   BrowserTokenRequest,
   BrowserTokenResponse,
   SeekApiBrowserTokenRequest,
-  SeekApiBrowserTokenResponse,
 } from './types';
 
-const tokenCache = new LRU<string, BrowserTokenResponse>();
+interface CacheItem {
+  expiry: string;
+  response: BrowserTokenResponse;
+}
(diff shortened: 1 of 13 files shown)


be/src/browserToken/middleware.ts
```diff
@@ -71,36 +92,40 @@ const _createBrowserTokenMiddleware = ({
...
-      authorization: `Bearer ${responseResult.value.access_token}`,
-      expiry: new Date(Date.now() + expiresInMs).toISOString(),
+    const freshItem: CacheItem = {
+      expiry,
+      response,
     };
 
     // Write off the token at its half life. This is a bit mean.
-    tokenCache.set(hirerId, freshToken, expiresInMs / 2);
+    tokenCache.set(cacheKey({ hirerId, scope }), freshItem, expiresInMs / 2);
```
etaoins, line 119: Do we still need to do this now that we have the "1 minute discount" added in this PR?

72636c: That's on the frontend cache; do you think the backend shouldn't cache at all then?

etaoins: No sorry, this is fair :+1:

PR #51, opened 2020-06-09
feat: Add LocationSuggest component

aaronvanston:
## Overview

Pulls in the LocationSuggest. A Location autocomplete component that can match on full words and substrings resolving a SEEK location. Returns a Location with object identifier on location selection.

This component can be used to fetch a location or used inline within a form that generates the input structure for other queries and mutations. For example, you can use this component to resolve a location OID for CategorySuggest or PositionPosting. 

## Includes:
- Storybook example, only resolves a mock query with Melbourne and Sydney location
- Readme with usage instructions.


fe/lib/hooks/Debounce/Debounce.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * @see {@link https://dev.to/gabe_ragland/debouncing-with-react-hooks-jci}
+ */
+export const useDebounce = <T>(value: T, delay: number) => {
```
72636c, line 6: Reckon we should just use something like https://github.com/xnimorz/use-debounce?

fe/package.json
```diff
@@ -1,8 +1,14 @@
 {
   "dependencies": {
+    "@apollo/react-hooks": "^3.1.5",
     "@types/uuid": "^8.0.0",
     "@types/yup": "^0.29.2",
+    "apollo-boost": "^0.4.9",
+    "autosuggest-highlight": "^3.1.1",
     "content-disposition": "^0.5.3",
+    "graphql": "^15.0.0",
+    "graphql-tools": "^6.0.7",
```
72636c, line 10: Is this a `devDependency`?

fe/package.json
```diff
@@ -1,8 +1,14 @@
 {
   "dependencies": {
+    "@apollo/react-hooks": "^3.1.5",
     "@types/uuid": "^8.0.0",
     "@types/yup": "^0.29.2",
+    "apollo-boost": "^0.4.9",
```
72636c, line 6: I'm keen on trying out the [Apollo Client 3 RC](https://github.com/apollographql/apollo-client/blob/v3.0.0-rc.2/CHANGELOG.md) to see if it resolves out caching woes, but that can be for another day 🙂

fe/lib/components/LocationSuggest/queries.ts
```diff
@@ -0,0 +1,66 @@
...
+    parent {
+      ...locationAttributes
+      parent {
+        ...locationAttributes
+        parent {
+          ...locationAttributes
+          parent {
+            ...locationAttributes
+            parent {
+              ...locationAttributes
```
72636c, line 26: Just wondering, is there any reasoning behind the level we go to here?

fe/lib/components/LocationSuggest/LocationSuggest.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { LocationSuggest } from './LocationSuggest';
+import { mockLocationSuggest } from './__fixtures__/locationSuggest';
+import { LOCATION_SUGGEST } from './queries';
+
+const mockClient = createMockClient();
+
+mockClient.setRequestHandler(LOCATION_SUGGEST, () =>
+  Promise.resolve({ data: mockLocationSuggest }),
+);
```
72636c, line 17: 👏

fe/lib/components/LocationSuggest/LocationSuggest.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+storiesOf('Locations', module)
+  .add('Locations Suggest', () => (
+    <LocationSuggest id="locationSuggest" schemeId="seekAnz" />
+  ))
+  .addDecorator((story) => (
+    <ApolloProvider client={mockClient}>
+      <BraidLoadableProvider themeName="seekAnz">
```
72636c, line 25: Should we use `seekUnifiedBeta`? 🤔

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ) => {
+    const [
+      getLocationSuggest,
+      { data: suggestData, error: suggestError },
+    ] = useLazyQuery(LOCATION_SUGGEST, {
+      ...(client && { client }),
+      fetchPolicy: 'no-cache',
+    });
+
+    const [selectedLocationId, setselectedLocationId] = useState('');
```
72636c, line 46:
```diff
- setselectedLocationId
+ setSelectedLocationId
```

fe/lib/components/LocationSuggest/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Required
+
+- `schemeId`: The scheme for the location dataset to query.
+
+Optional:
+
+- `client`: An `ApolloClient` instance. By default locationSuggest uses the client passed down via context, but a different client can be passed in.
+- `usageTypeCode`: The context that the location suggestions will be used for. Defaults to `PositionPosting`.
+- `hirerId`: The hirer identifier for relevant locations result.
```
72636c, line 25: Maybe something like:

```suggestion
- `hirerId`: The hirer identifier; enables tailored suggestions.
```

fe/lib/components/LocationSuggest/queries.ts
aaronvanston: No major rationale behind this. For a location of "Melbourne" It has 4 nested parents that it can resolve. here it's an arbitrary amount to cover our bases, I can reduce it down to 4/ the deepest known relation

> Melbourne > CBD & Inner Suburbs > Melbourne >  Victoria > Australia

72636c: Nah this is fine, I was just wondering whether we were tracking some standard around subdivision levels that we could document

fe/lib/components/LocationSuggest/LocationSuggest.stories.tsx
aaronvanston: Yeah makes sense! We can switch the others over to this theme too!

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -0,0 +1,119 @@
+import { useLazyQuery } from '@apollo/react-hooks';
+import ApolloClient from 'apollo-client';
+import { FieldMessage } from 'braid-design-system';
+import { FieldProps } from 'braid-design-system/lib/components/private/Field/Field';
```
72636c, line 4: I think the preferred pattern for Braid is to use something like `React.ComponentProps<typeof TextField>`

PR #52, opened 2020-06-09
refactor: rename QuestionnaireBuilder storybook stories 

aaronvanston:
Aligns structure of QuestionnaireBuilder stories to the current component structure

### DIFF
--- a/fe/lib/components/QuestionnaireBuilder/app.stories.tsx
+++ b/fe/lib/components/QuestionnaireBuilder/app.stories.tsx
@@ -6,9 +6,9 @@ import { storiesOf } from 'sku/@storybook/react';
 
 import { QuestionnaireBuilder } from './QuestionnaireBuilder/QuestionnaireBuilder';
 
-storiesOf('App', module)
-  .add('seekAnz', () => <QuestionnaireBuilder />)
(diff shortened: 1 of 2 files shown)


fe/lib/components/QuestionnaireBuilder/app.stories.tsx
```diff
@@ -6,9 +6,9 @@ import { storiesOf } from 'sku/@storybook/react';
 
 import { QuestionnaireBuilder } from './QuestionnaireBuilder/QuestionnaireBuilder';
 
-storiesOf('App', module)
-  .add('seekAnz', () => <QuestionnaireBuilder />)
-  .addDecorator((story) => (
+storiesOf('QuestionnaireBuilder', module)
+  .add('Builder', () => <QuestionnaireBuilder />)
+  .addDecorator(story => (
```
72636c, line 11: It seems that IDE integrations aren't picking up the Prettier config correctly 😢

I'll try to have a squiz at why later

aaronvanston: Yeah :/ I pushed the fix from running the `sku format`, was going to look into it also.

PR #67, opened 2020-06-24
feat: add JobCategorySuggest component

aaronvanston:
# Added
A JobCategory suggest component that provides a narrowed down list of SEEK job categories, these are ordered by confidence of suggestion.

Returns a `JobCategorySuggestionChoice` on selection of a suggestion. This includes a SEEK JobCategory with job category object identifier and a floating-point confidence score.

# Changed
- Remove unused `gql` queries for previous job category select

# Questions
- I've gone about common queries for Job Categories just by duplicating the structure/fragments. Is this fine? Should we abstract these out and share them among components?


fe/lib/components/JobCategorySuggest/JobCategorySuggest.tsx
```diff
@@ -0,0 +1,89 @@
...
+        getCategorySuggestion({
+          variables: {
+            positionProfile: debounceJobCategorySuggestInput,
+            schemeId,
+            first: 5,
+          },
+        });
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [schemeId, debounceJobCategorySuggestInput]);
```
aaronvanston, line 60: Same thing as before, listening to "getCategorySuggestion" function for updates caused extraneous renders.

72636c: I guess ESLint encourages us to memoise everything? Doesn't seem worth the effort.

aaronvanston: Yeah, I was going to look into eventually using `useMemo`, however at this stage, this was mostly a migration into wingman. I will eventually go back and update these setting a nicer pattern moving forward.

PR #70, opened 2020-06-29
Fix(JobCategorySuggest): field props not being passed through correctly 

aaronvanston:
# Overview
When using JobCategorySuggest as a field in a form you could not pass through props to the underlying fields

# Changes
- Removes hardcoded "name" prop on radio
- Creates a new field message to receive passed in "message" and "tone" 
- Pass tone, and spreads extra props (name and other field props) to the radio button

### DIFF
--- a/fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
+++ b/fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
@@ -1,5 +1,5 @@
 import { Radio, Stack, Strong, Text } from 'braid-design-system';
-import React, { forwardRef, useState } from 'react';
+import React, { ComponentPropsWithRef, forwardRef, useState } from 'react';
 
 import {
   JobCategory,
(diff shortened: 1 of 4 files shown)


fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -1,13 +1,15 @@
...
 import {
   JobCategory,
   JobCategorySuggestionChoice,
 } from '../../types/seek.graphql';
 import { flattenResourceByKey } from '../../utils';
 
-interface Props {
+interface Tone extends Pick<ComponentPropsWithRef<typeof Radio>, 'tone'> {}
+
+interface Props extends Tone {
```
72636c, line 12: Nit, but one of these may be preferable:

```suggestion
interface Props extends Tone {
  tone: ComponentProps<typeof Radio>['tone']
```

```suggestion
interface Tone extends Pick<ComponentPropsWithRef<typeof Radio>, 'tone'> {}

type Props = Pick<ComponentProps<typeof Radio>, 'tone'> & {
```

aaronvanston: Oh yeah, I like the first structure better. thanks! :)

PR #78, opened 2020-07-03
Fix questionnaire builder mutation variables

etaoins:
Right now we produce JSON with top-level `variables` and `valid` properties.  If you copy and paste this in to the Playground it will fail; you just need the variables.

This doesn't bother with `valid` as it doesn't seem to be possible to create an invalid questionnaire with the UI.


fe/lib/components/QuestionnaireBuilder/components/GraphqlQueryRenderer/GraphqlQueryRenderer.tsx
```diff
@@ -59,7 +59,7 @@ export const convertComponentsToMutationVariables = (
   hirerId: string,
 ): MutationVariables => {
   const questionnaireComponents: MutationQuestionnaireComponents[] = components.map(
-    (component) =>
+    component =>
```
72636c, line 62: Oh yeah Prettier does not get along well with this repo at the moment 😢

```suggestion
    (component) =>
```

PR #86, opened 2020-07-06
Add some routing to frontend example

72636c:


fe/example/src/App/App.treat.ts
```diff
@@ -0,0 +1,9 @@
+import { globalStyle } from 'sku/treat';
+
+globalStyle('body', {
+  height: '100vh',
+});
+
+globalStyle('#app, #app > div, #app > div > div', {
```
aaronvanston, line 7: This structure should not change, so I think it's fine. However, these nested styles could hurt us down the track. There should be ways to isolate/ attach the styles to the correct divs to push the height to 100vh.

72636c: Yeah this is gross, I'll have a look

72636c: That works a treat! Ryan always gets salty about 100vh breaking his mobile Safari though

etaoins: Mobile Safari is living in a hell it created for itself. Let it burn.

PR #95, opened 2020-07-08
Update to react-hook-form@6

etaoins:
- Built-in Yup support has been removed in favour of a `@hookform/resolvers` package that supports multiple schema validators.

  This just passes our existing Yup schema in to the new resolver.

- `react-hook-form` is moving away from `as` for non-trivial cases (see react-hook-form/react-hook-form#1580).

  Their types no longer include `onChange` for the `as` prop even though they're passing it. This works for most form elements but Braid requires `TextField`s to have an `onChange`.

  This seems to technically be a `react-hook-form` type bug but they seem to be pushing people towards the `render` prop anyway. I think the result also more clearly separates the form logic from the Braid logic.

### DIFF
--- a/fe/lib/components/QuestionnaireBuilder/QuestionnaireBuilder/FormBuilder/components/NewPrivacyConsentForm/NewPrivacyConsentForm.tsx
+++ b/fe/lib/components/QuestionnaireBuilder/QuestionnaireBuilder/FormBuilder/components/NewPrivacyConsentForm/NewPrivacyConsentForm.tsx
@@ -1,3 +1,4 @@
+import { yupResolver } from '@hookform/resolvers';
 import {
   Actions,
   Box,
@@ -44,7 +45,7 @@ export default ({
   const { dispatch } = useContext(StateContext);
 
   const { control, handleSubmit, errors } = useForm<FormValues>({
-    validationSchema: schema,
+    resolver: yupResolver(schema),
   });
 
   const saveThisPrivacyConsent = (formData: FormValues) => {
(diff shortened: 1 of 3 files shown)


fe/lib/components/QuestionnaireBuilder/QuestionnaireBuilder/FormBuilder/components/NewPrivacyConsentForm/NewPrivacyConsentForm.tsx
```diff
@@ -70,24 +71,32 @@ export default ({
             </Heading>
           </Box>
           <Controller
-            as={TextField}
+            render={(formProps) => (
```
72636c, line 74: I like this a lot more, it looks less magic and we probably get better typing out of it.

aaronvanston, line 74: Curious is the form props typed when you pull them off in this render function?

etaoins: Yes, `react-hook-form` is providing some loose types here:

https://github.com/react-hook-form/react-hook-form/blob/6a120e4165b7fdf724004e327b2c4dc87be31dd5/src/types/props.ts#L44-L48

PR #105, opened 2020-07-10
Expose questionnaire builder components from wingman-fe

lumishrestha:


fe/lib/components/index.ts
```diff
@@ -2,3 +2,20 @@ export { Example } from './Example/Example';
 export { JobCategorySelect } from './JobCategorySelect/JobCategorySelect';
 export { JobCategorySuggest } from './JobCategorySuggest/JobCategorySuggest';
 export { LocationSuggest } from './LocationSuggest/LocationSuggest';
+
+export { QuestionnaireBuilder } from './QuestionnaireBuilder/QuestionnaireBuilder/QuestionnaireBuilder';
+
+export { FormBuilder } from './QuestionnaireBuilder/QuestionnaireBuilder/FormBuilder/FormBuilder';
+export { FormRenderer } from './QuestionnaireBuilder/QuestionnaireBuilder/FormRenderer/FormRenderer';
```
72636c, line 9: Hmm, actually having second thoughts about some of these exports. I don't think these necessarily make sense in a shared package. We might need to namespace or otherwise rename them. e.g. this might be better as `QuestionnaireRenderer`, and the types as `Questionnaire.PrivacyConsent`

lumishrestha: So does `questionnaire.ts` have still have all the exports and it would be `Questionnaire.FormBuilder` or did you mean changing FormRender to QuestionnaireRenderer?

PR #107, opened 2020-07-12
Craft artisanal deployment

72636c:
This drops the Storybook deployer in favour of a simple script with plenty of hardcoding. This allows us to trivially customise the directory structure of the `gh-pages` branch and add in our mock site.

<https://seek-oss.github.io/wingman>

I initially thought of running a script after the Storybook deployer, but unfortunately it deletes its working directory after deployment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,13 +1,9 @@
 name: Release
 
-# TODO: restore
-# on:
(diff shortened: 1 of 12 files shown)


72636c:
Screen Shot 2020-07-13 at 01 26 18

PR #127, opened 2020-07-21
Add position list and detail pages

72636c:
The filter fields don't do anything, though they would be neat to implement.


72636c:
Screen Shot 2020-07-21 at 12 51 00
Screen Shot 2020-07-21 at 12 51 10
Screen Shot 2020-07-21 at 12 51 27
Screen Shot 2020-07-21 at 12 51 58

Ugh, the incorrectly-global local storage state of the ad form bites in the third screenshot.

PR #128, opened 2020-07-21
Add branding and questionnaire page stubs

72636c:
These should eventually become proper list views with edits, deletes and all that jazz.

They currently reuse the widgets that are embedded in the posting flow.


72636c:
Screen Shot 2020-07-21 at 13 09 58
Screen Shot 2020-07-21 at 13 12 25

PR #141, opened 2020-07-28
Add locations by geolocation

ak-08:
**Additions:** 
- Adds the ability to use geolocation to detect the location and suggest a location closest to the detected location. 
- Documentation on how the `nearestLocations` query is used within `LocationSuggest`.  

**Updates:** 
-  Wingman now uses the latest SEEK API generated types. 
- `JobCategorySuggest` now has `PositionLocation` as an additional input (required field) 

image


fe/lib/components/JobCategorySuggest/JobCategorySuggest.stories.tsx
```diff
@@ -23,6 +23,7 @@ storiesOf('JobCategories', module)
       schemeId="seekAnz"
       positionProfile={{
         positionTitle: `Senior Developer`,
+        positionLocation: ['seekAnzPublicTest:location:seek:2FqwWaaMV'],
```
aaronvanston, line 26: 👍  Thanks

fe/lib/components/LocationSuggest/LocationSuggest.stories.tsx
```diff
@@ -8,13 +8,17 @@ import { storiesOf } from 'sku/@storybook/react';
...
-import { LOCATION_SUGGEST } from './queries';
+import { mockNearestLocations } from './__fixtures__/nearestLocations';
+import { LOCATION_SUGGEST, NEAREST_LOCATIONS } from './queries';
 
 const mockClient = createMockClient();
 
 mockClient.setRequestHandler(LOCATION_SUGGEST, () =>
   Promise.resolve({ data: mockLocationSuggest }),
 );
+mockClient.setRequestHandler(NEAREST_LOCATIONS, () =>
```
aaronvanston, line 19: 👍

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -120,6 +176,13 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
             tone="critical"
           />
         )}
+        {nearestLocationsError && (
```
aaronvanston, line 179: In my opinion, I think we could refactor this to just use the existing `FieldMessage` for `suggestError`. Could combine them and have a general error: "Error fetching location, please try again". Both of these errors would be SEEK API Graph errors and fetching the location (either for lat/long or for autosuggest).

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -70,6 +87,24 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
       }
     };
 
+    const handleDetectLocationClicked = (input: GeoLocationInput | Error) => {
+      if (input instanceof Error) {
+        return (
+          <FieldMessage
```
aaronvanston, line 93: We should not return react component from this handle click function. There are a few options:
- We could either combine this error with the other graph errors
- Or specifically set-up state/ a new field message below for not being able to access the geolocation via navigator
- Or display the field error message within the `LocationSuggestInput component`

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -97,13 +132,34 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
     return (
       <Fragment>
         <LocationSuggestInput
+          isLoading={nearestLocationsLoading}
           onSelect={handleSuggestSelect}
-          onClear={() => setSelectedLocationId('')}
+          onDetectLocation={handleDetectLocationClicked}
+          onClear={() => {
+            setSelectedLocationId('');
+            setPlaceholder('');
```
aaronvanston, line 158: 👍

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
```diff
@@ -82,17 +99,48 @@ const LocationSuggestInput = ({
...
+        />
+      </Column>
+      <Column width="content">
+        <Button
+          loading={isLoading}
+          weight="weak"
+          onClick={handleDetectLocationClicked}
+        >
+          Detect location
+          <IconLocation />
```
aaronvanston, line 140: Something a bit picky, but we should add some padding between "Detect location" and the icon. You can wrap either in a Braid box and add padding left/right to it.

72636c: Is JSX eating the space in between because they aren't on the same line?

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -50,12 +54,25 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
     });
 
+    const [
+      nearestLocations,
+      {
+        data: nearestLocationsData,
+        error: nearestLocationsError,
+        loading: nearestLocationsLoading,
+      },
+    ] = useLazyQuery(NEAREST_LOCATIONS, {
```
72636c, line 64: Not for this PR, but it would be good to set up [TypeScript operation codegen](https://graphql-code-generator.com/docs/plugins/typescript-operations) so we can strongly type this like `useLazyQuery<NearestLocationsQuery, NearestLocationsQueryVariables>(...)`

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
aaronvanston: Yeah, another way to do this is @ak-08 is to add `{' '}` after "Detect location" to force it to render a space instead.

fe/lib/components/LocationSuggest/LocationSuggest.tsx
ak-08: Yeah sure, would you prefer doing something like 

`       {(suggestError|| nearestLocationsError) && (
          <FieldMessage
            id="suggestError"
            message="Error fetching location, please try again"
            tone="critical"
          />
        )}`

ak-08: ☝️ That merges both errors into a general error message thrown back

fe/lib/components/LocationSuggest/LocationSuggest.tsx
ak-08: >We should not return react component from this handle click function.

Any reason why not?

fe/lib/components/LocationSuggest/LocationSuggest.tsx
ak-08: I think these types are already available for us to use. 

I noticed this as well and some mocks also not being typed. I think it'll be good to clean them up in a follow-up.

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
ak-08: `Detect location <IconLocation />`

Doing this does the job 🥳

PR #145, opened 2020-07-30
Fix/nearest locations widget

ak-08:
**What**
- Type Location Queries and variables 
- Align 'Detect Location' button correctly 
- Fix bugs to detect location correctly

### DIFF
--- a/fe/lib/components/LocationSuggest/LocationSuggest.tsx
+++ b/fe/lib/components/LocationSuggest/LocationSuggest.tsx
@@ -132,7 +132,7 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
     }, [suggestData]);
 
     useEffect(() => {
-      if (nearestLocationsData && nearestLocationsData.nearestLocations) {
+      if (nearestLocationsData?.nearestLocations) {
         // The closest SEEK location returned for the geolocation input
         const closestLocation = nearestLocationsData.nearestLocations[0];
 
--- a/fe/lib/components/LocationSuggest/LocationSuggest.tsx
+++ b/fe/lib/components/LocationSuggest/LocationSuggest.tsx
@@ -13,6 +13,9 @@ import {
   GeoLocationInput,
   Location,
   LocationSuggestion,
+  Maybe,
+  QueryLocationSuggestionsArgs,
+  QueryNearestLocationsArgs,
 } from '../../types/seek.graphql';
 
 import LocationSuggestInput from './LocationSuggestInput';
@@ -48,7 +51,10 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
     const [
       getLocationSuggest,
       { data: suggestData, error: suggestError },
-    ] = useLazyQuery(LOCATION_SUGGEST, {
+    ] = useLazyQuery<
+      { locationSuggestions: Maybe<Array<LocationSuggestion>> },
+      QueryLocationSuggestionsArgs
+    >(LOCATION_SUGGEST, {
       ...(client && { client }),
       fetchPolicy: 'no-cache',
     });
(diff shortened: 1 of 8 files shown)


fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -132,25 +132,29 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
         setPlaceholder(contextualName);
-        setSelectedLocationId(value);
+
+        if (onSelect) {
+          onSelect(closestLocation);
+        }
+
+        if (closestLocation?.id?.value) {
+          setSelectedLocationId(closestLocation.id.value);
+        }
```
aaronvanston, line 149:
```suggestion
        handleLocationSelect(closestLocation)
```

You can remove this duplicate code and just use `handleLocationSelect`

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
```diff
@@ -44,7 +44,7 @@ interface Props {
   onSelect: (location?: Location) => void;
   onChange: (textInput: string) => void;
   onClear: () => void;
-  onDetectLocation: (input: GeoLocationInput | Error) => void;
```
aaronvanston, line 47: 👍

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -132,25 +132,29 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
     }, [suggestData]);
 
     useEffect(() => {
-      if (nearestLocationsData) {
+      if (nearestLocationsData && nearestLocationsData.nearestLocations) {
```
72636c, line 135:
```suggestion
      if (nearestLocationsData?.nearestLocations) {
```

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -60,7 +66,10 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
         error: nearestLocationsError,
         loading: nearestLocationsLoading,
       },
-    ] = useLazyQuery(NEAREST_LOCATIONS, {
+    ] = useLazyQuery<
+      { nearestLocations: Maybe<Array<Location>> },
```
72636c, line 70: AFAIK this isn't ideal because it isn't limited to the subset of fields you've selected in your query. This is what [TypeScript Operations](https://graphql-code-generator.com/docs/plugins/typescript-operations) will do, along with running your query against the SEEK API to ensure its validity

fe/lib/components/LocationSuggest/LocationSuggest.tsx
ak-08: Updated ✅

fe/lib/components/LocationSuggest/LocationSuggest.tsx
ak-08: Updated ✅

fe/lib/components/LocationSuggest/LocationSuggest.tsx
ak-08: Can you elaborate a bit more on this @72636c ?

72636c: If you have a schema like this:

```graphql
type Query {
  location: Location!
}

type Location {
  id: String!
  description: String!
}
```

A type generated from the schema won't necessarily reflect what you've selectively retrieved in your query:

```typescript
// Type generated from the schema
interface Location {
  id: string
  description: string
}

// Notice that we do not select `description`
const QUERY = gql`
  query Location {
    location {
      id
    }
  }
`;

const { data } = useQuery<{ location: Location }>(QUERY);

// This is invalid but TypeScript doesn't catch it
data.description;
```


The TypeScript Operations plugin can inspect `QUERY` and produce a better type like this:

```typescript
// Type generated from `QUERY`
interface LocationQuery {
  location: Pick<Location, 'id'>;
}
```

ak-08: Gothcya. That makes sense. Do we use this anywhere in Wingman yet so I can see how it works?

ak-08: 💥 Done! The types are dynamically typed using `typescript-operations`. We should do the same for `JobCategorySuggest` 

Tested on RA and everything looks ✅ 

@72636c

ak-08: P.S. I've just added the location queries as part of documents for now while generating types but we could do something similar to RA and nest it under `src/operations`

PR #170, opened 2020-08-14
Render questionnaire builder from GraphQL mutation

lumishrestha:
This adds a `FormBuilderQueryInput` component that takes in graphQL mutation variables and validates it. The component is exported on its own, so QuestionnaireBuilder has the option of using the variables from the component to create the questionnaire form.


fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -17,22 +17,33 @@ import {
...
 } from '../components/GraphqlQueryRenderer/GraphqlQueryRenderer';
-import { FormComponent } from '../questionTypes';
+import { mapMutationVariableToFormComponent } from '../mapping';
+import { FormComponent, MutationVariableInput } from '../questionTypes';
 
 import { FormBuilder } from './FormBuilder/FormBuilder';
 
 interface QuestionnaireBuilderProps {
   hirerId?: string;
+  graphqlInputVariables?: MutationVariableInput;
```
72636c, line 27: 1. It's confusing that there's a hirer ID nested in `graphqlInputVariables` that is ignored in favour of the separate `hirerId` prop. It might make sense to drop `hirerId`.
1. There no plural _variables_ here; how about just `graphqlInput`?

fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -17,22 +17,33 @@ import {
...
 }: QuestionnaireBuilderProps) => {
   const [formBuilderState, setFormBuilderState] = useState<FormComponent[]>([]);
   const [hirerId, setHirerId] = useState(externalHirerId ?? '');
 
+  useEffect(() => {
+    setFormBuilderState(
+      graphqlInputVariables
+        ? mapMutationVariableToFormComponent(graphqlInputVariables)
+        : [],
+    );
```
72636c, line 44: Should we make the effect conditional like:

```suggestion
    if (graphqlInputVariables) {
      setFormBuilderState(
        mapMutationVariableToFormComponent(graphqlInputVariables),
      );
    }
```

An undefined input doesn't seem valid so I'm not sure we should clear our state based on it.

fe/lib/components/Questionnaire/mapping.ts
```diff
@@ -38,3 +43,15 @@ export const mapApplicationQuestionnaireToFormComponent = (
...
+  graphqlInput: MutationVariableInput,
+): FormComponent[] => {
+  const components = graphqlInput.input.applicationQuestionnaire.components;
+
+  return components.map((component) =>
+    component.componentTypeCode === 'PrivacyConsent'
+      ? component.privacyConsent
+      : component.question,
+  );
+};
```
72636c, line 57: You can turn this into a simple expression

fe/lib/components/Questionnaire/index.ts
```diff
@@ -13,4 +13,5 @@ export {
   SelectionQuestion,
   FormComponent,
   PrivacyConsent,
+  MutationVariableInput,
```
72636c, line 16: Does this export do anything?

fe/lib/components/Questionnaire/QuestionnaireBuilder/FormBuilder/state/formBuilderState.ts
```diff
@@ -140,12 +151,16 @@ export const actionCreators = {
...
 } as const;
 
 interface StateContextType {
   state: FormComponent[];
   dispatch: React.Dispatch<
-    AddComponentAction | DeleteComponentAction | EditComponentAction
+    | AddComponentAction
+    | DeleteComponentAction
+    | EditComponentAction
+    | ImportComponentAction
```
72636c, line 163: I think this should be plural, like `ImportComponents/ReplaceComponents/SetComponents`

fe/lib/components/Questionnaire/components/FormBuilderQueryInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  return (
+    <ContentBlock>
+      <Stack space="gutter">
+        <Card>
+          <Stack space="gutter">
+            <Textarea
+              label="Graphql Variables"
+              id="id"
```
72636c, line 50: This doesn't seem right. It should probably be parameterised and provided by the consuming app.

fe/lib/components/Questionnaire/components/FormBuilderQueryInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+  };
+
+  return (
+    <ContentBlock>
+      <Stack space="gutter">
+        <Card>
+          <Stack space="gutter">
+            <Textarea
+              label="Graphql Variables"
```
72636c, line 49:
```suggestion
              label="GraphQL Input"
```

fe/lib/components/Questionnaire/components/FormBuilderQueryInput.tsx
lumishrestha: Did you mean the value for the text area?

72636c: This becomes a [global element ID in the DOM](https://developer.mozilla.org/en-US/docs/Web/API/Element/id) i.e. `<textarea id="id"></textarea>`, so it's not good for a reusable component to squat on a generic value like `id`.

fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
lumishrestha: 1. I think Ryanair gets the `hirerId` from `user`. Since `graphqlInput` is optional, there might be no hirerId? or should there be a check to see if hirerId is the same in the `graphqlInput`?
2. sure

fe/lib/components/Questionnaire/index.ts
lumishrestha: Ah, I exported it so ryanair could use the type for the variable. It seems `QuestionnaireCreateInput` is already exported and `MutationVariableInput` could just be derived from "variables" so maybe it's not needed.

fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
lumishrestha: Oh yep that makes sense

fe/lib/components/Questionnaire/index.ts
72636c: 👌 was just checking if this was important, as I don't think it gets re-exported out of the parent `fe/lib/components`

fe/lib/components/Questionnaire/questionTypes.ts
```diff
@@ -38,3 +50,71 @@ export interface PrivacyConsent {
...
+  privacyConsent: PrivacyConsent,
+}).Or(
+  RTRecord({
+    componentTypeCode: Literal('Question'),
+    question: FreeTextQuestion.Or(SelectionQuestion),
+  }),
+);
+
+const QuestionnaireMutationVariableInput = RTRecord({
+  input: RTRecord({
```
72636c, line 107: Could/should we drop this `input` nesting? It doesn't seem to do much.

lumishrestha: Can we leave it as it is? just to follow query variables syntax for the mutation

PR #177, opened 2020-08-18
Unnest graphQL input component

lumishrestha:
Removes unnecessary layers of braid so the consuming app can have a better layout


etaoins (approved):
👌

72636c (approved):

PR #193, opened 2020-09-02
Remove hirer ID from the questionnaire builder

etaoins:
This is currently broken unless a mutation is provided. Instead, this depends on our consumers supplying their own ID. This can either be provided from their session state or from an explicit input.

### DIFF
--- a/fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -16,14 +16,14 @@ import {
   QuestionnaireCreateInput,
   convertComponentsToMutationVariables,
 } from '../components/GraphqlQueryRenderer/GraphqlQueryRenderer';
-import { mapMutationVariableToFormComponent } from '../mapping';
-import { FormComponent, MutationVariableInput } from '../questionTypes';
+import { mapGraphqlToFormComponent } from '../mapping';
+import { FormComponent, GraphqlComponentInput } from '../questionTypes';
 
 import { FormBuilder } from './FormBuilder/FormBuilder';
 
 interface QuestionnaireBuilderProps {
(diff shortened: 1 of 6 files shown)


fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -23,23 +22,21 @@ import { FormComponent, MutationVariableInput } from '../questionTypes';
 import { FormBuilder } from './FormBuilder/FormBuilder';
 
 interface QuestionnaireBuilderProps {
+  hirerId: string;
   graphqlInput?: MutationVariableInput;
```
72636c, line 26: Unfortunately this is a bit messy; `graphqlInput` also has a hirer ID. Does this now ignore that ID?

etaoins: Yeah, I was thinking the caller would have to pull that out themselves. The other option would be to have disjoint prop types for `graphqlInput` and `hirerId`

72636c: I think I made @lumishrestha suffer enough to know to avoid the potential complexity of the latter.

Would it make sense to take in just the `components` array rather than the whole input?

72636c (approved):

fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
etaoins: How about d1718d5?

72636c: 😍

72636c (approved):

PR #204, opened 2020-10-05
Enable CodeQL GitHub code scanning

raphsutti:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,68 @@
...
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+name: 'CodeQL'
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 13 * * 6'
```
72636c, line 12: Looks like the schedule is randomised across the repos at the moment. Do you think that makes sense, or would it be better for us to get all the notifications at once?

72636c (approved):

.github/workflows/codeql-analysis.yml
raphsutti: A specific scan day/time make sense 🤔  
What day makes sense? Something like Every Monday at 9am?

72636c: Sounds good 👍

.github/workflows/codeql-analysis.yml
```diff
@@ -9,7 +9,7 @@ on:
   push:
   pull_request:
   schedule:
-    - cron: '0 13 * * 6'
+    - cron: '0 9 * * 1'
```
72636c, line 12: Is this in UTC? 😛

72636c (approved):

.github/workflows/codeql-analysis.yml
raphsutti: Timezones 🤦  yes it is 
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows

72636c (approved):
🚀 thanks for this

PR #241, opened 2020-11-05
fix: job category suggest

aaronvanston:
With #239 we changed the radio group value to be index based, as a result we need to update the managed radio group with the selected index not the SEEK ID.

- This changes the current useState used for job category to be purely index (int) based and not store the JobCategory choice. Within the `onSelect` passed in function we still send back the full jobCategory to the parent.


72636c (approved):
Ah d'oh, 🙇

PR #299, opened 2020-12-17
Uplift SpecifiedPersonForm

72636c:
- Use titlecased role labels
- Support initial values
- Only show required message on blur or submit
- Make email address mandatory

---

Kapture 2020-12-18 at 08 43 58


fe/lib/hooks/Fields/Fields.ts
```diff
@@ -0,0 +1,126 @@
+import { FormEvent, useState } from 'react';
+
+const transformValue = <K extends string, V>(
```
etaoins, line 3: Is this a monoid in the category of endofunctors?

etaoins (approved):

PR #337, opened 2021-02-09
Use Braid's TextLinkButton

72636c:
TextLinkRenderer is deprecated. We don't seem to lose much with this switch apart from the un-Braid-y `disabled` prop.


fe/lib/components/Questionnaire/QuestionnaireBuilder/FormBuilder/FormBuilder.tsx
```diff
@@ -135,7 +141,7 @@ export const FormBuilder = ({
       state.length < MAX_NUMBER_OF_COMPONENTS && (
         <Card key="privacyConsentCard">
           <Text>
-            <TextLinkButton onClick={onClickAddPrivacyConsent} width="full">
```
72636c, line 138: This doesn't appear to have regressed visually in Storybook 🤷

etaoins (approved):

PR #349, opened 2021-02-23
Pin `@hookform/resolvers` 1.3.0

72636c:
Really loving the negative value that Yup has added to my life.


72636c:
Great it's not even type compatible. Going to 🔥 this entirely.

PR #401, opened 2021-05-03
Clean up job category GraphQL queries

etaoins:
- Use typed GraphQL operations

- Don't query our children's parent. This is unneeded and increases the complexity of our GraphQL operation

- Inline the `nestedJobCategoryAttributes` fragements only used in a single place

- Type our props to only include the fields we're selecting, not the entire infinite `JobCategory` object

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -2,9 +2,7 @@ generates:
   ./fe/lib/types/seek.graphql.ts:
     schema: ${SEEK_SCHEMA_PATH:https://graphql.seek.com/graphql}
     documents:
-      - fe/lib/components/JobCategorySelect/queries.ts
-      - fe/lib/components/JobCategorySuggest/queries.ts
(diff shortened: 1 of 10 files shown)


codegen.yml
```diff
@@ -2,6 +2,8 @@ generates:
   ./fe/lib/types/seek.graphql.ts:
     schema: ${SEEK_SCHEMA_PATH:https://graphql.seek.com/graphql}
     documents:
+      - fe/lib/components/JobCategorySelect/queries.ts
+      - fe/lib/components/JobCategorySuggest/queries.ts
       - fe/lib/components/LocationSuggest/queries.ts
```
72636c, line 7: Should we wildcard this to e.g.

```suggestion
      - fe/lib/components/**/queries.ts
```

72636c (approved):

PR #419, opened 2021-05-19
Remove custom flexbox usage

72636c:
This is all better Braided through the dedicated column components.


fe/lib/components/Questionnaire/QuestionnaireBuilder/FormBuilder/components/NewQuestionForm/components/SelectOptions.tsx
```diff
@@ -65,26 +64,23 @@ export default ({ options, setOptionList }: SelectOptionsProps) => {
 
   return (
     <Stack space="medium">
-      <form onSubmit={submitHandler}>
```
72636c, line 68: This `form` wasn't used (notice that the `Button` below duplicates the listener rather than specifying `type=submit`) and should be avoided since this is nested inside another form.

etaoins (approved):

PR #420, opened 2021-05-19
Ensure question options are unique and non-empty

72636c:
This threads through a simplified setter signature so we aren't tied to supporting the flexibility of `useState` setters. This makes it easier for us to wire up `react-hook-form` state management.

Kapture 2021-05-19 at 12 55 49


fe/lib/components/Questionnaire/QuestionnaireBuilder/FormBuilder/components/NewQuestionForm/components/QuestionField.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 
 interface QuestionFieldProps {
   value: string;
-  setValue: React.Dispatch<React.SetStateAction<string>>;
```
etaoins, line 6: :+1:

etaoins (approved):

PR #484, opened 2021-07-22
Add Other option to job category choices

AshwaryaSmridhi:
**What is this change:**
This PR adds a `Other` option to the list of suggested job categories. 

**Why do we need this change:**
This is so RyanAir create flow follows UX recommendations and shows a full list of categories if the suggested list doesn't contain the category the hirer wants.

**Note:**
- We have added JobCategorySelect in the JobCategorySuggest component as suggested [here](https://github.com/seek-oss/wingman/pull/484#issuecomment-885247390).

Current UI:
image

### DIFF
--- a/fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
+++ b/fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
@@ -62,7 +62,7 @@ const JobCategorySelectInput = ({
         <Dropdown
           {...restProps}
           id="jobCategoriesSelect"
-          label={hideLabel ? "" : "Category"}
+          label={hideLabel ? undefined : 'Category'}
           onChange={(event) =>
             setSelectedParentCategoryId(event.currentTarget.value)
           }
@@ -81,7 +81,7 @@ const JobCategorySelectInput = ({
           <Dropdown
             {...restProps}
             id="subJobCategoriesSelect"
-            label={hideLabel ? "" : "Subcategory"}
+            label={hideLabel ? undefined : 'Subcategory'}
             onChange={(event) =>
               setSelectedChildCategoryId(event.currentTarget.value)
             }
--- a/fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
+++ b/fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
@@ -14,11 +14,13 @@ interface Props {
   jobCategories: AllJobCategories;
   onSelect: (jobCategory: AnyJobCategory, type: 'parent' | 'child') => void;
   tone: ComponentProps<typeof Dropdown>['tone'];
+  hideLabel?: boolean;
 }
 
 const JobCategorySelectInput = ({
   jobCategories,
   onSelect,
+  hideLabel,
   ...restProps
 }: Props) => {
   const [selectedParentCategoryId, setSelectedParentCategoryId] = useState('');
@@ -60,7 +62,7 @@ const JobCategorySelectInput = ({
         <Dropdown
           {...restProps}
           id="jobCategoriesSelect"
-          label="Category"
+          label={hideLabel ? "" : "Category"}
           onChange={(event) =>
             setSelectedParentCategoryId(event.currentTarget.value)
           }
@@ -79,7 +81,7 @@ const JobCategorySelectInput = ({
           <Dropdown
             {...restProps}
             id="subJobCategoriesSelect"
-            label="Subcategory"
+            label={hideLabel ? "" : "Subcategory"}
             onChange={(event) =>
               setSelectedChildCategoryId(event.currentTarget.value)
             }
--- a/fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
+++ b/fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
@@ -53,12 +53,6 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
       confidence: choice.confidence,
     }));
 
-    suggestions.push({
-      key: 'Other',
-      label: 'Other',
-      confidence: 0,
-    });
-
     const [selectedJobCategory, setSelectedJobCategory] = useState<string>();
 
     const handleChoiceSelect = (event: React.FormEvent<HTMLInputElement>) => {
@@ -78,7 +72,6 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
     return (
       <Stack space="small">
         <Text weight="strong">Category</Text>
-
         <RadioGroup
           {...restProps}
           name={name}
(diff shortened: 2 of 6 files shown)


AshwaryaSmridhi:
This PR is still a WIP. We have asked for review to just get some suggestions on the following:
- We still need to add the job category select list. We are still deciding if we should add that as a part of job category suggest list in wingman or add it in RyanAir. What are people's thoughts?

72636c:
Oops, forgot to submit. My 2c is we should aim to encapsulate complexity in the JobCategorySuggest component where practical; it's finicky to duct tape the components together in Ryanair, and ideally each component would present our best practice out of the box rather than requiring a bunch of additional work. LocationSuggest is a good example of this as it seamlessly bundles the (optional) location detection feature.

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -96,6 +101,24 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
             );
           })}
         </RadioGroup>
+        {selectedJobCategory === 'Other' && (
```
72636c, line 104: Can `JobCategorySelect` be nested as a child of `<RadioItem label="Other" />` above?

See https://seek-oss.github.io/braid-design-system/components/RadioGroup#toggling-nested-content.

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -96,6 +101,24 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
...
+          <JobCategorySelect
+            client={client}
+            name={name}
+            id="jobCategoryId"
```
72636c, line 108:
```suggestion
            id="job-category-suggest-select-other"
```

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -96,6 +101,24 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
...
+            onSelect={(jobCategory: JobCategoryAttributesFragment) => {
```
72636c, line 109: Is this type annotation required? Ideally the `JobCategorySelect` component would have its types sorted out itself.

fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
```diff
@@ -60,7 +62,7 @@ const JobCategorySelectInput = ({
         <Dropdown
           {...restProps}
           id="jobCategoriesSelect"
-          label="Category"
+          label={hideLabel ? "" : "Category"}
```
72636c, line 65:
```suggestion
          label={hideLabel ? undefined : 'Category'}
```

fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
```diff
@@ -79,7 +81,7 @@ const JobCategorySelectInput = ({
           <Dropdown
             {...restProps}
             id="subJobCategoriesSelect"
-            label="Subcategory"
+            label={hideLabel ? "" : "Subcategory"}
```
72636c, line 84:
```suggestion
            label={hideLabel ? undefined : 'Subcategory'}
```

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -73,7 +78,7 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
         <RadioGroup
           {...restProps}
           id="job-category-suggest-select"
-          name={name}
+          name={selectedJobCategory === 'Other' ? '' : name}
```
72636c, line 81: Hmm I'm not getting the point of this change, can you explain?

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -96,6 +101,24 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
...
+              const jobCategorySuggest = choices.find(
+                (choice) => choice.jobCategory.id.value === jobCategory.id.value,
+              );
+
+              if (onSelect && jobCategorySuggest) {
+                onSelect(jobCategorySuggest);
+              }
+            }}
+            schemeId={schemeId}
+            hideLabel={true}
```
72636c, line 119:
```suggestion
            hideLabel
```

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -96,6 +101,24 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
...
         </RadioGroup>
+        {selectedJobCategory === 'Other' && (
+          <JobCategorySelect
+            client={client}
+            name={name}
+            id="jobCategoryId"
+            onSelect={(jobCategory: JobCategoryAttributesFragment) => {
+              const jobCategorySuggest = choices.find(
+                (choice) => choice.jobCategory.id.value === jobCategory.id.value,
+              );
```
72636c, line 112: The `JobCategorySelect` component lets you select _any_ category, while `choices` is a subset that only contains a few top suggestions based on user input. You need to propagate the `jobCategory` parameter by itself to `onSelect`.

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
AkiraJ48: Sorry I didn't realise there were comments! I think the reason we had to do that was because I wasn't sure which value would be taken from which HTML element on form submit 😅  i.e. if I let both the radiogroup and select components have the `name`. (I'm assuming that's how `react-hook-forms` is pulling out input value?)

72636c: ohh I see, happy to give this a shot then

AkiraJ48: Hmm actually after delving deeper into the `react-hook-form` docs I may be wrong, I'll verify this lol

AkiraJ48: So the good news is that I'm pretty sure we don't need that, as we're treating these as controlled components. However I have noticed that the component is in a bit of weird state where I'm unsure of whether we wanted to allow it to be both uncontrolled/controlled. 

Right now `JobCategorySuggest` won't actually work if you don't pass it a `onSelect` prop (despite it being optional), so I'm thinking of making that a required prop and just asserting that `JobCategorySuggest` should be a controlled component. The main reasons are that the underlying `RadioGroup` component expects to be controlled as well as this new feature won't work without it being controlled.

72636c: Sounds good!

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -44,43 +61,61 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
...
+            label={'Other'}
+            ref={forwardedRef}
+            value={'Other'}
+          >
+            {selectedJobCategory === 'Other' && (
+              <JobCategorySelect
+                client={client}
+                id="job-category-suggest-select-other"
+                onSelect={(jobCategory) => {
+                  onSelect({ jobCategory, confidence: 1 });
```
etaoins, line 112: Note that the suggest will only return child categories, but `JobCategorySelect` will return both children and parent. The parent categories aren't suitable for job posting so this would break Ryanair's posting flow.

We can either:

- Pass through the `JobCategoryType` we're getting in this `onSelect` so the consumer (e.g. Ryanair) can ignore `parent` categories.

- Filter out `onSelect` for child categories here completely

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -44,43 +61,61 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
...
+                  )}
+                </RadioItem>
+              );
+            })}
+          </>
+          <RadioItem
+            key={'Other'}
+            label={'Other'}
+            ref={forwardedRef}
+            value={'Other'}
```
72636c, line 105:
```suggestion
            key="Other"
            label="Other"
            ref={forwardedRef}
            value="Other"
```

72636c (approved):
🤩

etaoins (approved):
🙇

PR #491, opened 2021-08-02
Remove `ssr` flag in graphql queries

AkiraJ48:
The expectation with this change is that all consumers will need to wrap the components with `ClientOnly` (which Ryanair is already doing).

The `ssr` flag should only be set when we're actually doing server side rendering (which we're not). This conflicts with the latest update by apollo client.


72636c (approved):
🙇

PR #506, opened 2021-08-13
Make questionnaire builder's GraphQL dump optional

etaoins:
This lets us hide our raw GraphQL output from the questionnaire builder.

This is useful for when the questionnaire builder is embedded in a job posting flow versus being used for developer documentation.

### DIFF
--- a/fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
+++ b/fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
@@ -25,12 +25,20 @@ interface QuestionnaireBuilderProps {
   hirerId: string;
   graphqlInput?: GraphqlComponentInput[];
   onChange?: (mutationVariables: QuestionnaireCreateInput) => void;
+
+  /**
+   * Controls if the raw GraphQL mutation & variables should be shown
+   *
+   * Defaults to `true`
+   */
+  showGraphqlOutput?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
```diff
@@ -65,16 +73,18 @@ export const QuestionnaireBuilder = ({
           </Column>
         </Columns>
 
-        <Card>
-          <Stack space="large">
-            <Heading level="3">GraphQL Output</Heading>
+        {showGraphqlOutput !== false && (
```
raphsutti, line 76:
```suggestion
        {showGraphqlOutput && (
```
Will checking truthy-ness work here?

etaoins: It's optional so that will make it default to `false` and change current behaviour.

Another option is having a `hideGraphqlOutput` property that works like a normal boolean but negative prop names are also annoying 🤔

raphsutti (approved):

fe/lib/components/Questionnaire/QuestionnaireBuilder/QuestionnaireBuilder.tsx
raphsutti: Ah I see - the property is optional; when the property exists, the value defaults to true

PR #509, opened 2021-08-15
Initial location with pre-filled value

AshwaryaSmridhi:
This PR is triggered by the piece of work to make editing better in RyanAir. 
We need the Location Suggest component to load the initial value.

### DIFF
--- a/.changeset/prefill-location-suggest.md
+++ b/.changeset/prefill-location-suggest.md
@@ -1,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+**LocationSuggestInput:** Allow `LocationSuggestInput` to load initial location
--- a/.changeset/prefill-location-suggest.md
+++ b/.changeset/prefill-location-suggest.md
@@ -2,4 +2,4 @@
 'wingman-fe': minor
 ---
 
-**LocationSuggestInput:** Allow `LocationSuggestInput` to load initial location
+**LocationSuggestInput:** Support loading an initial location
--- a/fe/lib/components/LocationSuggest/LocationSuggest.tsx
+++ b/fe/lib/components/LocationSuggest/LocationSuggest.tsx
@@ -146,7 +146,11 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
     ]);
 
     const loadInitialLocation = useCallback(async () => {
-      if (!initialValue || initialLocation?.id.value === initialValue) {
+      if (
+        !initialValue ||
+        initialLocation?.id.value === initialValue ||
+        selectedLocationId
+      ) {
         return;
       }
 
--- a/fe/lib/components/LocationSuggest/LocationSuggest.tsx
+++ b/fe/lib/components/LocationSuggest/LocationSuggest.tsx
@@ -12,6 +12,8 @@ import { useDebounce } from 'use-debounce';
 import type {
   GeoLocationInput,
   Location,
+  LocationQuery,
+  LocationQueryVariables,
   LocationSuggestion,
   NearestLocationsQuery,
   NearestLocationsQueryVariables,
@@ -153,7 +155,10 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
       if (!client) {
         return;
       }
-      const { data } = await client.query({
+      const { data } = await client.query<
+      LocationQuery,
+      LocationQueryVariables
+    >({
         ...(client && { client }),
         fetchPolicy: 'no-cache',
         query: LOCATION,
--- a/fe/lib/components/LocationSuggest/LocationSuggest.tsx
+++ b/fe/lib/components/LocationSuggest/LocationSuggest.tsx
@@ -100,9 +100,7 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
     const [detectLocationError, setDetectLocationError] = useState<string>();
 
     const [placeholder, setPlaceholder] = useState('');
-    const [initialLocation, setInitialLocation] = useState<
-      Location | undefined
-    >();
+    const [initialLocation, setInitialLocation] = useState<Location>();
     const [debounceLocationSuggestInput] = useDebounce(
       locationSuggestInput,
       debounceDelay,
@@ -156,9 +154,9 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
         return;
       }
       const { data } = await client.query<
-      LocationQuery,
-      LocationQueryVariables
-    >({
+        LocationQuery,
+        LocationQueryVariables
+      >({
         ...(client && { client }),
         fetchPolicy: 'no-cache',
         query: LOCATION,
@@ -178,7 +176,7 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
 
     useEffect(() => {
       loadInitialLocation();
-    });
+    }, [loadInitialLocation]);
 
     useEffect(() => {
       if (suggestData?.locationSuggestions) {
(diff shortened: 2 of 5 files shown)


fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -139,6 +154,23 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
+        });
+      }
+    }, [initialValue, getLocation]);
+
+    useEffect(() => {
+      setInitialLocation(locationData?.location ?? undefined);
+      if (locationData?.location) {
+        setSelectedLocationId(locationData?.location.id.value);
+      }
+    }, [locationData]);
```
72636c, line 172: Can/should we collapse this into one effect? e.g.

```typescript
const loadInitialLocation = useCallback(async () => {
  if (!initialValue || initialLocation?.id.value === initialValue) {
    return;
  }

  const { data } = await client.query();

  if (!data.location) {
    return;
  }

  setInitialLocation(data.location);

  // TODO: do we want this if?
  // It would avoid overriding the user's selection if they are somehow super quick.
  if (!selectedLocationId) {
    setSelectedLocationId(data.location.id.value);
  }
}, [initialValue]);

useEffect(() => {
  loadInitialLocation();
}, [blah]);
```

.changeset/prefill-location-suggest.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+**LocationSuggestInput:** Allow `LocationSuggestInput` to load initial location
```
72636c, line 5:
```suggestion
**LocationSuggestInput:** Support loading an initial location
```

IreneLee0905 (approved):

phongtcao (approved):

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -139,6 +145,36 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
 
+    const loadInitialLocation = useCallback(async () => {
+      if (!initialValue || initialLocation?.id.value === initialValue) {
+        return;
+      }
+
+      if (!client) {
+        return;
+      }
+      const { data } = await client.query({
```
72636c, line 156: I think specifying the types here is good for safety:

```suggestion
      const { data } = await client.query<
        LocationQuery,
        LocationQueryVariables
      >(
```

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -139,6 +145,36 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
     ]);
 
+    const loadInitialLocation = useCallback(async () => {
+      if (!initialValue || initialLocation?.id.value === initialValue) {
+        return;
+      }
+
+      if (!client) {
+        return;
+      }
```
72636c, line 155: D'oh, I didn't realise this was optional! I think we should make `client` a required prop now, otherwise it'll be confusing when we hit this code path and the feature mysteriously doesn't work.

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -139,6 +145,36 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
+      setInitialLocation(data.location);
+
+      if (!selectedLocationId) {
+        setSelectedLocationId(data.location.id.value);
+      }
+    }, [initialValue, client, selectedLocationId, initialLocation]);
+
+    useEffect(() => {
+      loadInitialLocation();
+    });
```
72636c, line 176: This will prevent the effect from re-running after the initial render:

```suggestion
    }, []);
```

Linting may ask for this which is fine too:


```suggestion
    }, [loadInitialLocation]);
```

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -95,6 +98,9 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
     const [detectLocationError, setDetectLocationError] = useState<string>();
 
     const [placeholder, setPlaceholder] = useState('');
+    const [initialLocation, setInitialLocation] = useState<
+      Location | undefined
```
72636c, line 102: FWIW you can actually omit `| undefined` as `useState` can infer it from the initial state

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -139,6 +145,36 @@ export const LocationSuggest = forwardRef<HTMLInputElement, Props>(
...
+        fetchPolicy: 'no-cache',
+        query: LOCATION,
+        variables: { id: initialValue },
+      });
+
+      if (!data.location) {
+        return;
+      }
+
+      setInitialLocation(data.location);
```
72636c, line 167: Hmm, I'm now thinking my suggestion may have been misguided. What do you think about this:

1. Scrap `initialLocation` and `setInitialLocation` altogether. I think we only needed that state to plumb between the two effects previously.
2. Change earlier guard clause from `initialLocation?.id.value === initialValue` to `selectedLocationId`. If the user has already made a selection I think the initial location no longer matters and we shouldn't need to run the query at all.

The function as-is is going to rerun the query on every `selectedLocationId` change, which is not ideal.

72636c (approved):
👏

fe/lib/components/LocationSuggest/LocationSuggest.tsx
AshwaryaSmridhi: HI Ryan,

I am passing `initialLocation` to the component. Are you saying don't pass the whole location object and just pass the id?

```
<LocationSuggestInput
            {...restProps}
            isLoading={nearestLocationsLoading}
            locationSuggestions={locationSuggestResults}
            onChange={setLocationSuggestInput}
            onClear={() => {
              setSelectedLocationId('');
              setPlaceholder('');
            }}
            onDetectLocation={handleDetectLocationClicked}
            onSelect={handleLocationSelect}
            placeholder={placeholder}
            setDetectLocationError={setDetectLocationError}
            tone={tone}
            initialLocation={initialLocation}
          />
          ```

Comment (72636c) [reply fe/lib/components/LocationSuggest/LocationSuggest.tsx:170]: Ah my bad, I see you're propagating the contextual name too. I think we just need to extend the initial guard clause then:

```diff
- if (!initialValue || initialLocation?.id.value === initialValue) {
+ if (!initialValue || initialLocation?.id.value === initialValue || selectedLocationId) {
  return;
}

File: fe/lib/components/LocationSuggest/LocationSuggest.tsx (hunk 4 above)
Comment (AshwaryaSmridhi) [reply fe/lib/components/LocationSuggest/LocationSuggest.tsx]: Thank you for re-writing the story part to make this work. Do we also need to remove the default usage scenario from here: https://github.com/seek-oss/wingman/tree/master/fe/lib/components/LocationSuggest#default-usage

Maybe make Usage with Apollo Client the default scenario?

Comment (72636c) [reply fe/lib/components/LocationSuggest/LocationSuggest.tsx]: I think we can update the default usage example with `useApolloClient`, which is doing the same thing as is described in the comment:

> // Higher component in tree wraps children in apollo provider

File: fe/lib/components/LocationSuggest/README.md [hunk 8]
```diff
@@ -43,9 +43,12 @@ Extends:
 
 ```javascript
 import { LocationSuggest } from 'wingman-fe';
+import { useApolloClient } from '@apollo/client';
+
+const client = useApolloClient();
 
 // Higher component in tree wraps children in apollo provider
-<LocationSuggest schemeId="seekAnz" />;
+<LocationSuggest schemeId="seekAnz" client={client} />;
```
AshwaryaSmridhi, line 51: @72636c ⬆️

PR #512, opened 2021-08-16
Prefill Job category

AshwaryaSmridhi:
This PR will automatically fill job category parent and child dropdown.

To do
- [x] Pass down parent and child IDs as a prop
- [x] Auto select the initial value for parent and child categories

image


fe/lib/components/JobCategorySelect/JobCategorySelect.tsx
```diff
@@ -27,6 +33,7 @@ interface Props extends Omit<FieldProps, 'value' | 'onChange' | 'children'> {
     type: JobCategoryType,
   ) => void;
   schemeId: string;
+  initialValue?: { parent?: string | undefined; child?: string | undefined };
```
etaoins, line 36: I'm a bit suspect of this interface:

- What happens if the `parent` & `child` don't match?
- This pushes complexity on the caller to look up the correct `parent` & `child`

Would it be better to have an `initialJobCategoryId` and then figure out if it's a parent or child category internally?

phongtcao: yeah I think the caller (i.e `PositionProfileCreator`) only has a child JobCategoryId (unless we include parent in the PositionProfile query), in addtion, we also query list of all categories in `JobCategorySelect`, we would only need one input for a selected category Id and can look up from there

AshwaryaSmridhi: So we would pass down the `initialJobCategoryId` which is essentially child category Id and then do a look up to find the right parent Id? 

I did go down that path in this commit: https://github.com/seek-oss/wingman/pull/512/commits/6395f03ae9e48c81880bde28cdb7af34327d0e42 (Please ignore the horrible nested for loop. I always intended to replace that)

etaoins: You should be able to determine if `initialJobCategoryId` is a parent or child once you look it up in the list of job categories. From there you can handle it appropriately.

AshwaryaSmridhi: @etaoins and @phongtcao let me know if this solves your concern. 

> From there you can handle it appropriately - either just set a parent, or look up a child’s parent.

@etaoins from what I can see we will always pass down child id as job category id. Did you want us to solve for a scenario where only parent id is passed down? In that case we just set the parent id and keep the child selection empty?

etaoins: It'd be good to support parent IDs as well. This a common open source component that Ryanair happens to use; we should make it generalisable to other use cases.

fe/lib/components/JobCategorySelect/JobCategorySelect.tsx
```diff
@@ -58,6 +66,39 @@ export const JobCategorySelect = forwardRef<HTMLInputElement, Props>(
...
+
+    const loadInitialJobCategory = useCallback(async () => {
+      if (!initialValue || initialJobCategory?.id.value === initialValue) {
+        return;
+      }
+
+      if (!client) {
+        return;
+      }
+      const { data } = await client.query({
```
etaoins, line 81: Do we need to query the SEEK API again? Aren't we already querying the entire hierarchy with `JOB_CATEGORIES`?

etaoins: Would something like this work in `JobCategorySelectInput` instead:

```typescript
useEffect(() => {
  if (initialJobCategoryId && selectedParentCategoryId === '') {
    for (const parentCategory of jobCategories) {
      if (!parentCategory.children) {
        // Shouldn't happen but TypeScript is suspicious
        continue;
      }

      if (parentCategory.id === initialJobCategoryId) {
        // Passed a parent category ID
        setChildCategories(parentCategory.children);
        setSelectedParentCategoryId(parentCategory.id.value);
        return;
      }

      for (const childCategory of parentCategory.children) {
        if (childCategory.id === initialJobCategoryId) {
          setChildCategories(parentCategory.children);
          setSelectedParentCategoryId(parentCategory.id.value);
          setSelectedChildCategoryId(parentCategory.id.value);
          return;
        }
      }
    }
  }
}, [initialJobCategoryId, jobCategories, selectedParentCategoryId]);
```

AshwaryaSmridhi: Hmm, I was trying to avoid a nested loop. But nested loop vs querying the api, it makes sense to have the nested loop.

fe/lib/components/JobCategorySelect/queries.ts
```diff
@@ -19,3 +19,18 @@ export const JOB_CATEGORIES = gql`
   }
   ${JOB_CATEGORY_ATTRIBUTES}
 `;
+
+export const JOB_CATEGORY = gql`
```
etaoins, line 23: Can this be removed now?

fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
```diff
@@ -15,20 +15,23 @@ interface Props {
   onSelect: (jobCategory: AnyJobCategory, type: 'parent' | 'child') => void;
   tone: ComponentProps<typeof Dropdown>['tone'];
   hideLabel?: boolean;
+  initialJobCategoryId?: string;
```
etaoins, line 18: Maybe this should be `initialValue` as well for consistency

fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
```diff
@@ -15,20 +15,23 @@ interface Props {
...
   ...restProps
 }: Props) => {
   const [selectedParentCategoryId, setSelectedParentCategoryId] = useState('');
   const [selectedChildCategoryId, setSelectedChildCategoryId] = useState('');
   const [childCategories, setChildCategories] = useState<AnyJobCategory[]>();
+  const [isInitialJobCategory, setIsInitialJobCategory] = useState<boolean>();
 
   useEffect(() => {
-    if (selectedParentCategoryId !== '') {
+    if (selectedParentCategoryId !== '' && !isInitialJobCategory) {
```
etaoins, line 34: Do we know why these are `useEffect`s instead of just being in the `onChange` handlers? That would avoid the weird `initialJobCategory` state

AshwaryaSmridhi: It is on initial load. onChange would get triggered when we select a different option.

etaoins: Right, but do we *want* this to run on initial load? It seems like the `isInitialJobCategory` is just there to suppress it

AshwaryaSmridhi: Hmm, good point. Let me try that

AshwaryaSmridhi: Good pick up. There was no need of that useEffect.

fe/lib/components/JobCategorySelect/JobCategorySelectInput.tsx
```diff
@@ -15,33 +15,56 @@ interface Props {
...
+            return;
+          }
+        }
       }
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [selectedParentCategoryId]);
+  }, [initialValue, jobCategories]);
 
   useEffect(() => {
```
etaoins, line 69: Can this be a handler as well for consistency? AFAICS it has the same behaviour as the parent handler. It'd also be good to group them in the source file instead of having the `initialValue` handling splitting them up.

AshwaryaSmridhi: Also good point. I ignored that useEffect 🤦

fe/lib/utils/findObjectByChildOid.ts
```diff
@@ -0,0 +1,12 @@
+export const findObjectByChildOid = (objects: any, id: string) => {
```
etaoins, line 1: This looks to be unused now

fe/lib/components/JobCategorySuggest/JobCategorySuggestChoices.tsx
```diff
@@ -68,6 +70,17 @@ const JobCategorySuggestChoices = forwardRef<HTMLInputElement, Props>(
       }
     };
 
+    if (initialValue && selectedJobCategory === undefined) {
+      setSelectedJobCategory('Other');
+
+      const jobCategorySuggest = choices.find(
```
etaoins, line 76: Did we want to `onSelect` here? I don't think that's consistent with the other controls taking an initial value. If not, we should be able to set "Other" directly on line 57.

fe/lib/utils/findObjectByChildOid.ts
AshwaryaSmridhi: 😱

etaoins (approved):
🙇

AshwaryaSmridhi:
@etaoins We don't need to change documentation for this PR.

72636c (approved):
🏆

Definitely interested to see how `JobCategorySuggest` loads with this given it still requires enough `positionProfile` inputs to be able to retrieve suggestions before it renders any choices, including other.

Maybe this isn't an issue in practice when editing since the `positionProfile` should be available from the get-go.

AshwaryaSmridhi:
So we always have the positionProfile otherwise there should not be an initial value for this either.

PR #522, opened 2021-08-19
Drop requirement for Node.js `Buffer` polyfill

72636c:
This caused the questionnaire builder to WSoD when creating a new question under a Webpack 5 environment like sku 11.


fe/lib/components/Questionnaire/QuestionnaireForm/components/componentId.ts
```diff
@@ -1,5 +1,13 @@
 export const componentId = (prefix: string, ...rawValues: string[]) => {
-  const values = rawValues.map((value) => Buffer.from(value).toString('hex'));
+  const values = rawValues.map((str) => {
+    let value = '';
+
+    for (let i = 0; i < str.length; i++) {
+      value += str.charCodeAt(i).toString(36);
```
72636c, line 6: `toString(16)` would be equivalent to the prior code, but figured I'd bump this while we're here

etaoins (approved):
👏

PR #526, opened 2021-08-19
Uplift Storybooking

72636c:
- Migrate to Component Story Format

  https://storybook.js.org/docs/react/api/csf/

- Migrate to Storybook Controls

  https://storybook.js.org/docs/react/essentials/controls

- Rationalise story grouping and naming

  https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy

- Exclude tests from npm package

---

Screen Shot 2021-08-20 at 12 11 48 am


fe/lib/.npmignore
```diff
@@ -0,0 +1,9 @@
+/storybook/**/*
```
etaoins, line 1: Is this just for package size reasons?

I was reading the Storybook v6 documentation and I'm starting to see a method in their madness. We could *almost* automatically lint links to our Storybook from the Developer Site without taking any Storybook-specific deps. We'd just need to collect all the named exports from the `.stories.tsx` files.

`withBraidProvider` takes a dep on knobs but maybe we could move that to a global decorator so it's not explicitly imported.

etaoins (approved):

fe/lib/.npmignore
72636c: Yep, and it's not huge savings. I'll re-include the stories.

PR #531, opened 2021-08-20
Reorganise frontend components

72636c:
The directory structure here made it a bit confusing to jump from a top-level exported component to its source, which could be deeply tucked away in a non-obvious spot. We also had some exports that were not obviously named according to their corresponding feature.


.changeset/empty-singers-heal.md
```diff
@@ -0,0 +1,13 @@
+---
+'wingman-fe': minor
```
etaoins, line 2: Isn't this technically a breaking change?
```suggestion
'wingman-fe': major
```

etaoins (approved):

.changeset/empty-singers-heal.md
72636c: We've been cheating semver by staying on 0.x, but it's probably time for us to take things seriously.

PR #535, opened 2021-08-20
Add `Mock.showStorybookAction` prop

72636c:
Screen Shot 2021-08-21 at 2 19 08 pm


fe/lib/components/JobCategorySelect/JobCategorySelect.mock.tsx
```diff
@@ -1,21 +1,34 @@
...
+    <ApolloMockProvider
+      resolvers={{
+        Query: {
+          jobCategories: () => mockJobCategories,
+        },
+      }}
+    >
+      <MockComponentActions
+        space="medium"
+        storybookPath={
```
etaoins, line 24: Nice one! This is much better than hardcoding `base64` mystery URLs by hand.

etaoins (approved):

PR #539, opened 2021-08-23
Uplift QuestionnaireBuilder

72636c:
This was motivated by excess Card padding when we placed the QuestionnaireBuilder inside a Dialog as part of our new inlined posting flow. There were a few niggles that have built up over time that I was keen to address here.

Kapture 2021-08-23 at 15 00 35


etaoins (approved):
Until I read `CustomWrapper` I thought the GIF was somehow using Miro's change highlighting 🤦

PR #566, opened 2021-09-02
Add additional Apollo type policies

etaoins:
This is based on a review of objects that have an queryable primary object identifier. I was avoiding parts of the schema only accessible with a partner token but Apollo client can also be used server-to-server.

### DIFF
--- a/.changeset/late-dancers-cross.md
+++ b/.changeset/late-dancers-cross.md
@@ -4,4 +4,4 @@
 
 **apolloTypePolicies:** Add additional Apollo type policies
 
-This adds a policy `ApplicationQuestionnaire` and objects that are only accessible with a partner token.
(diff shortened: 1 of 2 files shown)


.changeset/late-dancers-cross.md
```diff
@@ -0,0 +1,7 @@
+---
+'wingman-fe': patch
+---
+
+**apolloTypePolicies:** Add additional Apollo type policies
+
+This adds a policy `ApplicationQuestionnaire` and objects that are only accessible with a partner token.
```
72636c, line 7:
```suggestion
This adds a policy for `ApplicationQuestionnaire` and objects that are only accessible with a partner token.
```

72636c (approved):

PR #571, opened 2021-09-04
Add `acceptedScopes` browser token option

etaoins:
This allows the backend to filter out the scopes it will request from the SEEK API.

### DIFF
--- a/be/README.md
+++ b/be/README.md
@@ -106,6 +106,13 @@ const createApp = () => {
   const middleware = createBrowserTokenMiddleware({
     getPartnerToken,
     userAgent: 'my-service/1.2.3',
+
+    acceptedScopes: [
+      'query:advertisement-brandings',
+      'query:ad-products',
+      'query:ontologies',
+      'query:organizations',
+    ],
   });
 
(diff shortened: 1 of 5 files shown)


be/README.md
```diff
@@ -106,6 +106,13 @@ const createApp = () => {
   const middleware = createBrowserTokenMiddleware({
     getPartnerToken,
```
72636c, line 107: Since we're on a breaking change tear at the moment, I think `getPartnerToken` may be better as something like `authorize` that returns a list of allowed scopes for the requester. This way we can properly enforce different roles and potentially open up read-only access to more internal users.

There's another decision in whether we:

1. Remove scopes from the request and have the middleware simply request all scopes returned by the `authorize` function for the user. The frontend can then use that token wherever it needs.
2. Continue to request specific scopes and have the middleware validate those against the allowed scopes for the user before passing them on to the SEEK API.

72636c: For multiple account support, it may also make sense for the request to specify the `hirerId` that the user is acting on behalf of.

PR #576, opened 2021-09-08
Remove Brandings sample page

phongtcao:
As we're gonna have a `BrandSelect` component in `wingman-fe`, we don't need a sample page for Brandings


72636c (approved):
🙇

PR #600, opened 2021-09-22
Fix linking line being hidden for job category select

AkiraJ48:
The linking line between the parent and child category were being hidden behind the card in Ryanair due to its `z-index`. I've gone and removed that and instead just calculated the correct height and width for it, as well as set the curvature to only occur to the bottom left of the border.


72636c (approved):
😌

PR #609, opened 2021-09-28
Add GitHub links to mock component actions

etaoins:
The existence of Wingman is still a bit mysterious. This adds an explicit link from our `MockComponentActions` to the relevant directory on GitHub. This will allow curious developers to go directly from the mock component on the Developer Site to the Wingman source.

Screen Shot 2021-09-28 at 14 25 18

PR #611, opened 2021-09-28
Add interactive salary component

AkiraJ48:
This adds an interactive salary component that will be added to Ryanair and the developer site.

This isn't a 1:1 mapping to the UX guide. Some key differences are:
- Removed "Hide pay information on your ad" checkbox" (I talked with Angelica and she agreed that it was confusing, for now I've opted to remove it)
- Changed "Customise the pay information displayed on your ad" to "Pay shown on your ad" which I believe is the wording used in direct
- ~~The little character limit is slightly different as I'm leveraging Braids `characterLimit` associated to the `TextField`~~
- I've also added the currency next to "Pay range"

I'll talk with Angelica again and see if she's okay with this, and then we can also update the UX guide. I also opted for the component to only take an `onBlur` atm as I'd prefer to not have the parent component have to handle constant state updates due to user typing. (I'm also not too happy with how I did the validation, so happy for any suggestions)

Screen Shot 2021-09-28 at 4 53 03 pm

### DIFF
--- a/fe/lib/components/SalaryDetails/README.md
+++ b/fe/lib/components/SalaryDetails/README.md
@@ -1,0 +1,32 @@
+# SalaryDetails
+
+Add details here
+
+## Installation
+
+```shell
+$ yarn add wingman-fe
+```
+
+### Properties
+
+Required
+
+- `onChange`: The callback to handle the completed specified person input. // make this onBlur basically (prevents spamming handlers)
+
+### Usage
+
+```javascript
+import React, { useState } from 'react';
+import { SpecifiedPersonForm } from 'wingman-fe';
+
+const AddSpecifiedPersonForm = () => {
+  const [person, setPerson] = useState();
+  return (
+    <div>
+      <SpecifiedPersonForm onCreate={setPerson} />
+      Your {person.roleCode}'s name is {jobCategory.name}
+    </div>
+  );
+};
+```
--- a/fe/lib/components/SalaryDetails/README.md
+++ b/fe/lib/components/SalaryDetails/README.md
@@ -21,7 +21,7 @@ Optional:
 - `initialMaximumPay`: A default value to initially render for the maximum pay.
 - `initialSalaryType`: A default value to initially render for the salary type.
 - `initialSalaryDescription`: A default value to initially render for the salary description.
-- `errors`: An object of errors mapping to either `payType`, `minimumPay`, `maximumPay` or `salaryDescription` that contain an error message.
+- `errors`: An object of errors mapping to either `salaryType`, `minimumPay`, `maximumPay` or `salaryDescription` that contain an error message.
 
 ### Validation
 
@@ -39,7 +39,7 @@ const PostingForm = () => {
 
   const onBlur = (item) => {
     switch (item.key) {
-      case 'payType':
+      case 'salaryType':
         setJobAd({ ...jobAd, [item.key]: item.type })
         return;
       case 'minimumPay':
--- a/fe/lib/components/SalaryDetails/README.md
+++ b/fe/lib/components/SalaryDetails/README.md
@@ -1,6 +1,6 @@
 # SalaryDetails
 
-Add details here
+A Salary Details component that contains information on the pay type, range and description.
 
 ## Installation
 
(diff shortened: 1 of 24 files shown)


fe/lib/components/SalaryDetails/SalaryDetails.mock.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { SalaryDetails, SalaryDetailsProps } from './SalaryDetails';
+
+interface Props extends SalaryDetailsProps {
+  showStorybookAction?: boolean;
+}
+
+export const MockSalaryDetails = ({ showStorybookAction, ...props }: Props) => (
+  <MockComponentActions
+    space="medium"
+    storybookPath={
```
etaoins, line 14: Can we update this for #609?

fe/lib/components/SalaryDetails/README.md
```diff
@@ -0,0 +1,66 @@
...
+### Usage
+
+```javascript
+import React, { useState } from 'react';
+import { SpecifiedPersonForm } from 'wingman-fe';
+
+const PostingForm = () => {
+  const [jobAd, setJobAd] = useState();
+
+  const onBlur = (item) => {
```
etaoins, line 43: This is pretty far away from the input the SEEK API actually expects:
<https://developer.seek.com/schema/#/named-type/RemunerationPackageInput>

Should we align the interface of the component more to the SEEK API, or give an example of the correct mapping in the README?

AkiraJ48: Yea that's a good point, I'll align some of the terminology so that it's closer to what the SEEK API expects e.g. `salaryType` to `basisCode` and then update the README so it constructs an object matching `RemunerationPackageInput`

etaoins (approved):
🚀

AkiraJ48:
Just waiting on Angelica to give an okay with the design, then I'll merge

fe/lib/index.ts
```diff
@@ -19,6 +19,8 @@ export { QuestionnaireForm } from './components/QuestionnaireForm/QuestionnaireF
 export { QuestionnaireQueryInput } from './components/QuestionnaireQueryInput/QuestionnaireQueryInput';
 export { QuestionnaireQueryOutput } from './components/QuestionnaireQueryOutput/QuestionnaireQueryOutput';
 
+export { SalaryDetails } from './components/SalaryDetails/SalaryDetails';
+export { MockSalaryDetails } from './components/SalaryDetails/SalaryDetails.mock';
```
phongtcao, line 23:
```suggestion
export { MockSalaryDetails } from './components/SalaryDetails/SalaryDetails.mock';

```

fe/lib/components/SalaryDetails/README.md
```diff
@@ -0,0 +1,105 @@
...
+    maximumAmount: renumeration.maximumAmount ?? {
+      currency: renumeration.currency,
+      value: Number(renumeration.maximumAmount),
+    },
+  }],
+  descriptions: [renumeration.description],
+})
+
+const PostingForm = () => {
+  const [renumeration, setRenumeration] = useState({
```
phongtcao, line 57: `renumeration` is misspelling
```suggestion
  const [remuneration, setRemuneration] = useState({
```

fe/lib/components/SalaryDetails/SalaryDetails.tsx
```diff
@@ -0,0 +1,165 @@
...
+          onBlur({ key: 'description', description: value })
+        }
+        onClear={() => onBlur({ key: 'description', description: '' })}
+        value={description}
+        placeholder={'Example content'}
+        tone={descriptionValidation.tone}
+        message={
+          descriptionValidation.message ?? 'e.g $50,000 + car + annual bonus'
+        }
+        characterLimit={MAX_CHAR_LIMIT}
```
phongtcao, line 161: just to clarify, I don't see limit in the [description](https://developer.seek.com/schema/#/named-type/RemunerationPackage/field/descriptions)

phongtcao (approved):

fe/lib/components/SalaryDetails/SalaryDetails.tsx
AkiraJ48: Oh yea that's a good point, I must've mistaken this for the hirer job reference 😅 , thanks for catching that, I'll remove that validation

etaoins: It looks like it *is* limited to 50 bytes:
<https://github.com/SEEK-Jobs/ie-position-posting-api/blob/2de31c458730e59a58cded1343fd473f32572857/src/http/controllers/helpers/validators/validateRemunerationDescriptions.ts#L1>

And the limit is documented on the input type:
<https://developer.seek.com/schema/#/named-type/RemunerationPackageInput/field/descriptions>

We usually don't document the limit on the output types because we don't enforce limits on the data we return, only what we accept.

AkiraJ48: Oh man, if only I had a waited another minute

phongtcao: ohh, sorry for confusing you by the wrong reference in schema :(

PR #617, opened 2021-09-29
Control GitHub links with `showStorybookAction`

etaoins:
When this is `false`/`undefined` we still show the "View on GitHub" link which is unlikely what the consumer wants. This repurposes the flag to control both the storybook & GitHub links as a group.

This also lets us tighten the types on our internal `MockComponentActions` to ensure the mock component is providing all the relevant links.

### DIFF
--- a/fe/lib/components/AdSelectionFallback/AdSelectionFallback.mock.tsx
+++ b/fe/lib/components/AdSelectionFallback/AdSelectionFallback.mock.tsx
@@ -15,10 +15,9 @@ export const MockAdSelectionFallback = forwardRef<HTMLSelectElement, Props>(
   ({ showStorybookAction, ...props }, ref) => (
     <MockComponentActions
       space="medium"
+      showStorybookAction={showStorybookAction}
       storybookPath={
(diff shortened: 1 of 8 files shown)


fe/lib/components/AdSelectionFallback/AdSelectionFallback.mock.tsx
```diff
@@ -15,10 +15,9 @@ export const MockAdSelectionFallback = forwardRef<HTMLSelectElement, Props>(
   ({ showStorybookAction, ...props }, ref) => (
     <MockComponentActions
       space="medium"
+      showStorybookAction={showStorybookAction}
       storybookPath={
-        showStorybookAction
-          ? '/story/job-posting-ad-selection-adselectionfallback--ad-selection-fallback'
-          : undefined
+        '/story/job-posting-ad-selection-adselectionfallback--ad-selection-fallback'
```
72636c, line 20: Can we use `storybookPath=""` syntax for these now?

72636c (approved):

PR #619, opened 2021-09-29
Increase spacing on `MockSalaryDetails` actions

etaoins:
This consists of a multiple controls which makes the `medium` spacing look a bit crowded.

### DIFF
--- a/.changeset/big-mice-camp.md
+++ b/.changeset/big-mice-camp.md
@@ -1,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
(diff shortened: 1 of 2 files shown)


etaoins:
## Before

Screen Shot 2021-09-29 at 18 06 24

## After

Screen Shot 2021-09-29 at 18 09 49

72636c (approved):
🤔 I wonder if large should be the default

etaoins:
It actually looks alright for locations & job categories IMO. We get in trouble here because it looks like it’s clustering the button with “Pay shown on your ad” without more spacing.

PR #621, opened 2021-09-30
Show `SpecifiedPerson` name components on same row

etaoins:
We have lots of horizontal space on desktop to display the names beside each other. This should free up some precious vertical space on Ryanair.

### DIFF
--- a/.changeset/purple-eggs-wave.md
+++ b/.changeset/purple-eggs-wave.md
@@ -1,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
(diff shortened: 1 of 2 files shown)


etaoins:
Not sure if UX has any thoughts here because it doesn't appear in our UX guide 🤔 

## Before
Screen Shot 2021-09-30 at 11 30 39

## After
Screen Shot 2021-09-30 at 11 30 47

72636c (approved):
I think this part is fine, but the dodgy contact list below defs needs some UX love, and Add contact button should probably be wrapped with `Actions`

PR #629, opened 2021-09-30
Set input type for SpecifiedPersonForm email/phone

etaoins:
This makes little difference on desktop browsers, but causes mobile software keyboards to customise their layout.


etaoins:
Simulator Screen Shot - iPhone 13 - 2021-09-30 at 16 08 25

Simulator Screen Shot - iPhone 13 - 2021-09-30 at 16 03 30

72636c (approved):

PR #644, opened 2021-11-03
Separate JobCategorySuggest positionProfile prop

griffithsh:
Using an object here was confusing React's equality checking into reloading the widget.

You're now required to pass two props (`positionTitle` and `positionLocation`) instead of one `positionProfile`.

### DIFF
--- a/fe/lib/components/JobCategorySuggest/JobCategorySuggest.tsx
+++ b/fe/lib/components/JobCategorySuggest/JobCategorySuggest.tsx
@@ -12,7 +12,6 @@ import { useDebounce } from 'use-debounce';
 import type {
   JobCategorySuggestQuery,
   JobCategorySuggestionChoiceAttributesFragment,
-  JobCategorySuggestionPositionProfileInput,
 } from '../../types/seekApi.graphql';
 
 import JobCategorySuggestChoices from './JobCategorySuggestChoices';
@@ -22,7 +21,8 @@ type RadioProps = ComponentPropsWithRef<typeof RadioGroup>;
 
 export interface JobCategorySuggestProps
   extends Partial<Omit<RadioProps, 'id'>> {
-  positionProfile: JobCategorySuggestionPositionProfileInput;
+  positionTitle: string;
+  positionLocation: string[];
   schemeId: string;
   onSelect: (
     jobCategorySuggestionChoice: JobCategorySuggestionChoiceAttributesFragment,
@@ -42,7 +42,8 @@ export const JobCategorySuggest = forwardRef<
       client,
       debounceDelay = 250,
       onSelect,
-      positionProfile,
+      positionTitle,
+      positionLocation,
       schemeId,
       showConfidence,
       initialValue,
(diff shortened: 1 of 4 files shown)


fe/lib/components/JobCategorySuggest/JobCategorySuggest.tsx
```diff
@@ -22,7 +21,8 @@ type RadioProps = ComponentPropsWithRef<typeof RadioGroup>;
 
 export interface JobCategorySuggestProps
   extends Partial<Omit<RadioProps, 'id'>> {
-  positionProfile: JobCategorySuggestionPositionProfileInput;
```
jonoc330, line 25: I'm guessing by not using this input anymore this will mean we'll no longer support passing in a list of [`positionFormattedDescriptions`](https://developer.seek.com/schema/#/named-type/JobCategorySuggestionPositionProfileInput/field/positionFormattedDescriptions) and [`positionOrganizations`](https://developer.seek.com/schema/#/named-type/JobCategorySuggestionPositionProfileInput/field/positionOrganizations). Are we comfortable with that?

griffithsh: Where were those being used?

jonoc330: Assuming this `JobCategorySuggestionPositionProfileInput` type matches our schema: https://developer.seek.com/schema/#/named-type/JobCategorySuggestionPositionProfileInput

etaoins: Ryanair uses it as well:
https://github.com/SEEK-Jobs/indie-ryanair-frontend/blob/73869d4002ea207475494db42540370bdc3d02d1/src/pages/Posting/jobCategories/components/JobCategorySuggestForm.tsx#L53-L69

griffithsh: Alright, going back to the original question, there are two issues to consider. One is that RyanAir will possibly get less accurate category suggestions with this new simplified interface, and the other is that partners looking to wingman for guidance may implement something that provides less accurate suggestions also. The two paths that I can see from here are to decide not worry about those issues and merge this as is, or add the missing fields to the interface as optional.

What do we reckon?

etaoins: This interface is the only way to use AIPS’s category predict interface. This is more of the Back Office having very basic needs than Ryanair/partners having special needs. We definitely need to keep supporting it.

The third option is to use `React.memo` with a custom equality function:
<https://reactjs.org/docs/react-api.html#reactmemo>

griffithsh: I don't like the idea of adding more props to this component, because I think it looks complex. Happy to explore React.memo deeper. Are you seeing that becoming part of the implementation of this JobCategorySuggest component, or back at the call site in the backoffice?

etaoins: We should place it in Wingman - it’s the components responsibility to keep its state sane when its inputs change

griffithsh:
Going to try an implementation that avoids the problem by memoizing its input instead.

PR #656, opened 2021-11-22
Don't rerender `JobCategorySuggest` on no-op props

etaoins:
This avoids re-querying the SEEK API and re-displaying suggestions unless there's a meaningful change to our props.


etaoins:
Best reviewed without whitespace:
<https://github.com/seek-oss/wingman/pull/656/files?diff=unified&w=1>

72636c (approved):
Beautiful. Would we be better off looping through an enumeration of the props rather than hardcoding them in this way? I'm just worried that a prop could be missed in future.

etaoins:
I'm actually missing the `RadioProps` already. I'll just use <https://github.com/FormidableLabs/react-fast-compare>.

72636c (approved):

PR #672, opened 2021-11-29
Redocument

griffithsh:
https://trello.com/c/TFl6rqmQ/1551-document-initialvalue-prop-of-jobcategorysuggest-component-from-wingman

These docs have slowly drifted out of date and needed a few changes to line them up with the latest JobCategorySuggest component.


griffithsh:
See https://github.com/seek-oss/wingman/pull/673 instead.

PR #693, opened 2021-12-20
Limit question response choices to 99

72636c:
You'd have to try really hard to do this in the UI, but this should also serve as documentation.

---

Example with the limit artificially reduced to 3:

| Below limit | At limit
| :-: | :-:
| Screen Shot 2021-12-21 at 8 00 47 am | Screen Shot 2021-12-21 at 8 01 17 am

### DIFF
--- a/fe/lib/components/QuestionnaireBuilder/FormBuilder/components/NewQuestionForm/components/SelectOptions.tsx
+++ b/fe/lib/components/QuestionnaireBuilder/FormBuilder/components/NewQuestionForm/components/SelectOptions.tsx
@@ -5,6 +5,7 @@ import {
   Columns,
   FieldMessage,
   Stack,
+  Text,
   TextField,
 } from 'braid-design-system';
 import React, { useMemo, useRef } from 'react';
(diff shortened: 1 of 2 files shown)


fe/lib/components/QuestionnaireBuilder/FormBuilder/components/NewQuestionForm/components/SelectOptions.tsx
```diff
@@ -99,39 +105,45 @@ export default ({ options, setOptionList }: SelectOptionsProps) => {
...
-                <TextField
-                  {...field}
-                  id="questionnaireBuilderAddOption"
-                  label="Options"
-                  onClear={() => setValue('option', '')}
-                  ref={optionInputRef}
-                  tone={errors.option ? 'critical' : undefined}
-                />
-              )}
+      {options.length >= RESPONSE_CHOICE_LIMIT ? (
```
raphsutti, line 108: Should this be `>`

raphsutti (approved):

fe/lib/components/QuestionnaireBuilder/FormBuilder/components/NewQuestionForm/components/SelectOptions.tsx
72636c: That would allow you to add `limit + 1` choices

PR #732, opened 2022-02-28
Refine messaging when there are no brands

72636c:


fe/lib/components/BrandSelect/BrandSelect.mock.tsx
```diff
@@ -11,12 +11,14 @@ import { mockBrands } from './__fixtures__/brands';
...
   showStorybookAction?: boolean;
   pageSize?: number;
+  variant?: 'multiple-brands' | 'no-brands';
 }
 
 export const MockBrandSelect = ({
   client: _client,
   showStorybookAction,
   pageSize,
+  variant = 'multiple-brands',
```
72636c, line 21: This will allow us to render the `no-brands` variant on the Developer Site

fe/lib/components/BrandSelect/PaginatedBrands.tsx
```diff
@@ -36,11 +36,11 @@ export const PaginatedBrands = ({
     return (
       <Notice tone="info">
         <Text>
-          You have no brands. Create a new one in the{' '}
+          You have no brands. Upload a new brand on the{' '}
           <SmartTextLink href="https://talent.seek.com.au/account/branding">
```
72636c, line 40: This hyperlink could present localisation challenges. Should we remove it?

fe/lib/components/BrandSelect/PaginatedBrands.tsx
```diff
@@ -36,11 +36,11 @@ export const PaginatedBrands = ({
...
-            SEEK AdCentre
-          </SmartTextLink>
-          .
+            SEEK employer website
```
72636c, line 41: "SEEK employer website" matches terminology on https://developer.seek.com/use-cases/job-posting/branding.

I'd appreciate wordsmithing input for this entire sentence though.

fe/lib/components/BrandSelect/PaginatedBrands.tsx
etaoins: That's an interesting point! I'd leave it in for now but something to think about - e.g. I wonder if we should have a URL redirection service for these long-lived URLs that partners might incorporate in to their software?

etaoins (approved):

phongtcao (approved):

PR #762, opened 2022-04-21
Trigger onSelect function for preselecting first brand

IreneLee0905:
Need to trigger the onSelect function to send the preselected brand id to the form.
> When you are on the position posting page, fill in required details, then select Standout for the ad type, the UI will preselect the first brand for you. This selected brand is not sent as part of the payload for posting a position or previewing the job ad.

### DIFF
--- a/.changeset/two-cherries-guess.md
+++ b/.changeset/two-cherries-guess.md
@@ -2,4 +2,4 @@
 'wingman-fe': patch
 ---
 
-Fix branding preselect
+BrandSelect: Trigger `onSelect` when preselecting the first brand
--- a/.changeset/two-cherries-guess.md
+++ b/.changeset/two-cherries-guess.md
@@ -1,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
+Fix branding preselect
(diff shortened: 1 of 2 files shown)


fe/lib/components/BrandSelect/BrandSelect.tsx
```diff
@@ -80,8 +80,12 @@ export const BrandSelect = ({
 
       // Preselect the first brand from the query.
       setSelectedBrandId(firstBrandId);
+
+      if (onSelect) {
+        onSelect(firstBrandId);
+      }
```
72636c, line 86: Are we allowed to do this:

```suggestion
      onSelect?.(firstBrandId);
```

72636c (approved):

72636c:
You'll want to add a changeset ^ too

72636c (approved):

.changeset/two-cherries-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
+Fix branding preselect
```
72636c, line 5:
```suggestion
BrandSelect: Trigger `onSelect` when preselecting the first brand
```

phongtcao (approved):

.changeset/two-cherries-guess.md
IreneLee0905: 🙇‍♀️

PR #794, opened 2022-06-14
Add breadcrumbs to location suggest

tadhglewis:


.changeset/pretty-pugs-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+Add breadcrumbs to LocationSuggest
```
72636c, line 5:
```suggestion
LocationSuggest: Add breadcrumbs
```

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -224,6 +226,14 @@ export const LocationSuggest = forwardRef<
             tone="critical"
           />
         ) : null}
+
+        {selectedLocation ? (
+          <BreadCrumbsString
```
72636c, line 231: I don't think we want this to display in all usages of location suggest; it likely does not need to be there in an actual posting form or the example on the Developer Site. Can the components be decoupled or can this be an opt-in feature?

tadhglewis: Added an optional toggle. We could also separate add this to developer dashboard and ryanair but I think this is common enough to add into wingman

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -35,6 +37,7 @@ export interface LocationSuggestProps
   schemeId: string;
   usageTypeCode?: string;
   initialValue?: string;
+  showBreadcrumbs?: boolean;
```
72636c, line 40: Does this need to be added here: https://github.com/seek-oss/wingman/blob/95d6662691ce3b6cb29783da4a185b560aa50c99/fe/lib/components/LocationSuggest/LocationSuggest.stories.tsx#L21

.changeset/pretty-pugs-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+LocationSuggest: Add breadcrumbs
```
72636c, line 5:
```suggestion
LocationSuggest: Support `showBreadcrumbs` prop
```

fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -182,7 +186,7 @@ export const LocationSuggest = forwardRef<
           <input
             type="hidden"
             name={name}
-            value={selectedLocationId}
+            value={selectedLocation?.id.value}
```
72636c, line 189: Providing an `undefined` value to a form element is not desirable as it toggles the component between controlled and uncontrolled.

tadhglewis: Fixed

fe/lib/components/LocationSuggest/LocationSuggest.tsx
tadhglewis: No, it is optional. Storybook will just have a "set boolean" button, after clicking you can set true or false. If it's not set, it'll be the same as false.

I've added it anyway though just to be more explicit with the props

72636c (approved):
🙇

PR #796, opened 2022-06-15
Fix incorrect ordering of breadcrumbs

tadhglewis:
Screen Shot 2022-06-15 at 1 01 13 pm

Screen Shot 2022-06-15 at 1 01 26 pm

Screen Shot 2022-06-15 at 1 01 44 pm

### DIFF
--- a/.changeset/young-bees-smoke.md
+++ b/.changeset/young-bees-smoke.md
@@ -1,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
+LocationSuggest, LocationLookup, JobCategoryLookup: fix incorrect ordering of breadcrumbs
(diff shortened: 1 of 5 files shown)


tadhglewis:
Even with this change, the breadcrumbs are a bit weird when compared to traditional breadcrumbs.

This lists closest > broadest with closest being bold which is not like normal breadcrumbs - maybe it is fine as it's kinda more of a reverse breadcrumb to see where the closest crumb comes from

I think it'd also look weird to do broadest > closest with closest being bold

Screen Shot 2022-06-15 at 1 26 42 pm

tadhglewis:
Maybe we should reverse the arrows but keep **closest** < broadest

Screen Shot 2022-06-15 at 1 31 18 pm

.changeset/young-bees-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
+LocationSuggest, LocationLookup, JobCategoryLookup: fix incorrect ordering of breadcrumbs
```
72636c, line 5:
```suggestion
JobCategoryLookup, LocationLookup, LocationSuggest: Fix incorrect ordering of breadcrumbs
```

72636c (approved):

tadhglewis:
Done as per https://github.com/seek-oss/wingman/pull/796#issuecomment-1155944707 ... not a big fan of left arrows as they look weird but it'll do

PR #823, opened 2022-09-21
Drop `uuid` dep

72636c:
We're pretty aggressive on our browser policy and this appears to be supported by all major browsers.

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility


72636c:
Ah lame, looks like we may need TypeScript 4.6+ for this to pass type checking (seek-oss/sku#698)

PR #834, opened 2022-10-14
feat: Find nearest locations on map

tbolg:
This PR adds to the existing `LocationAutosuggest` component by adding a button to the component that opens a dialog to allow looking up locations on a map using the `nearestLocations` query.

New LocationAutosuggest appearance:
Screen Shot 2022-10-14 at 12 25 46 pm

Map dialog:
Screen Shot 2022-10-14 at 12 26 41 pm

Selecting a point overlays suggestions on the map: 
Screen Shot 2022-10-14 at 12 27 44 pm

Tablets and smaller devices show the suggested locations below the map: 
Screen Shot 2022-10-14 at 12 28 18 pm


fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
```diff
@@ -153,6 +174,41 @@ const LocationSuggestInput = ({
...
+      <Column width="content">
+        <Button
+          icon={<IconLanguage />}
+          onClick={() => setShowMapDialog(true)}
+          tone={tone === 'critical' ? tone : undefined}
+          variant="soft"
+        >
+          Select on map
+        </Button>
+        {showMapDialog ? (
```
AkiraJ48, line 186: Just wondering, do we still need this ternary?

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
```diff
@@ -153,6 +174,41 @@ const LocationSuggestInput = ({
...
+            }
+            open={showMapDialog}
+            onClose={() => {
+              setShowMapDialog(false);
+            }}
+          >
+            <MapSelect
+              schemeId={schemeId}
+              onLocationSelected={handleMapLocationSelected}
+              initialLocation={SEEK_MEL_CENTER}
```
AkiraJ48, line 207: I think it may make more sense to embed this default in the `MapSelect`, unless you were thinking of allowing consumers to override the initial location

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }>(NEAREST_LOCATIONS, {
+    variables: {
+      geoLocation: {
+        latitude: debouncedLatLong[0],
+        longitude: debouncedLatLong[1],
+      },
+      first: 3,
+      schemeId,
+    },
+    fetchPolicy: 'no-cache',
```
AkiraJ48, line 61: Is this still necessary with the new design?

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+    </Box>
+  );
+
+  const noLocationsNotice = () => (
+    <Alert
+      tone="info"
+      onClose={() => setShowSuggestions(false)}
+      closeLabel="Close"
+    >
+      <Text>{`No locations found.`}</Text>
```
AkiraJ48, line 124:
```suggestion
      <Text>No locations found.</Text>
```

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+        <Marker
+          color={color.foreground.formAccentLight}
+          anchor={latLong}
+          onClick={() => setCenter(latLong)}
+        />
+        {/* View for desktops and above */}
+        {showSuggestions && isDesktopOrAbove && (
+          <Overlay anchor={latLong} offset={suggestionsOffset}>
+            {hasNearestLocations
+              ? nearestLocationsCards()
```
AkiraJ48, line 155: It's typically preferable to use React syntax when rendering a component 🙂  e.g.
```suggestion
              ? <NearestLocationCards />
```

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+}
+
+export const MapSelect = ({
+  debounceDelay = 250,
+  schemeId,
+  onLocationSelected,
+  initialLocation,
+}: Props) => {
+  const [center, setCenter] = useState(initialLocation);
+  const [latLong, setLatLong] = useState(initialLocation);
```
AkiraJ48, line 43: I haven't dug into the code too deeply, but what's the difference between these 2? It seems like they could be used interchangeably

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const [showSuggestions, setShowSuggestions] = useState(false);
+
+  const responsivevalue = useResponsiveValue();
+  const color = useColor();
+  const [debouncedLatLong] = useDebounce(latLong, debounceDelay);
+
+  const { data, previousData } = useQuery<{
+    nearestLocations: Array<Location> | null;
```
72636c, line 51: Because GraphQL lets the client select a subset of fields, you'll generally want to use a type here which is generated based on your specified client operation rather than _all_ fields that are included in the schema. This should be codegenned automatically based on the `gql` query you have in your TypeScript code and there should be examples of this elsewhere in the codebase, but let me know if you get stuck on this one.

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
tbolg: Nope we don't, I'll remove it 👍

fe/lib/components/LocationSuggest/components/queries.ts
```diff
@@ -0,0 +1,15 @@
+import { gql } from '@apollo/client';
+
+export const NEAREST_LOCATIONS = gql`
+  query ($geoLocation: GeoLocationInput!, $schemeId: String!) {
```
72636c, line 4: To my point above, I think you will have to name this query in order to get a usable codegenned type. It looks like we have a mix of upper and lower camel cased names across this codebase 😬

72636c (commented):
This is wonderful!

fe/lib/components/LocationSuggest/components/MapSelect.tsx
tbolg: Nope, will remove 👍

fe/lib/components/LocationSuggest/components/MapSelect.tsx
tbolg: In the old design one was used as a state to keep track of and set the center of the map, whilst the other was used to control the position of the marker. This allowed you to create a marker at a single lat long, and then set the center separately, e.g. clicking on a marker brought the maps centre to it. In the new design I have it so that the map centres on each marker placement, meaning that we can just use one of these instead of both. I have removed the `centre` state and kept the `latLong` state 👍

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
tbolg: Yeah I did this with that idea in mind, e.g. might want to dynamically set the location depending on what region the user is viewing from. Probably ties into [this](https://github.com/seek-oss/wingman/pull/834#discussion_r996506967) comment below from @72636c

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+        </Box>
+        {nearestLocations?.map((location) => (
+          <Box
+            className={styles.card}
+            cursor="pointer"
+            padding="medium"
+            key={location.id.value}
+            onClick={() => onLocationSelected(location)}
+          >
+            <Box position="relative">
```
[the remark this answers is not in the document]
tbolg: Nope! Removed 😄

fe/lib/components/LocationSuggest/components/MapSelect.tsx
```diff
@@ -0,0 +1,176 @@
...
+const DEFAULT_ZOOM = 10;
+
+interface Props {
+  debounceDelay?: number;
+  schemeId: string;
+  onLocationSelected: (location: Location) => void;
+  initialLocation: [number, number];
+}
+
+export const MapSelect = ({
```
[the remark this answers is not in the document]
tbolg: Yep I think that would be good. I think having a separate display of it the developer dashboard like you had built is good to draw more direct attention to the `nearestLocations` query. Although having it as a part of the autosuggest is good, it is secondary to the autosuggest functionality being shown as the main focus in this component in the dev dashboard

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
72636c: Scope creep, but this now makes me wonder whether there should be some interplay with the "detect location" feature. I imagine it may make sense for `initialLocation` to start at the current location of the user.

fe/lib/components/LocationSuggest/components/MapSelect.tsx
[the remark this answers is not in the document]
tbolg: I have made this into a standalone `LocationSelectMap` component that the autosuggest now imports and uses. I also added a story for it so it can be viewed in isolation in the storybook.

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
tbolg: @72636c Agreed, I have made it so that if a detected location exists it will centre on that instead of SEEK HQ 😄

fe/lib/utils/formatPoint.ts
```diff
@@ -0,0 +1,7 @@
+export const formatPoint = (point: [number, number]): string => {
+  const roundedLat = Math.abs(Math.round(point[0] * 100) / 100);
+  const roundedLon = Math.abs(Math.round(point[1] * 100) / 100);
+  const lat = roundedLat < 0 ? `${roundedLat}° S` : `${roundedLat}° N`;
+  const lon = roundedLon < 0 ? `${roundedLon}° W` : `${roundedLon}° E`;
```
72636c, line 5: Is this logic correct? It seems like `roundedX < 0` will always return false given the `Math.abs`.

Would we rather want something like this

```suggestion
export const formatPoint = (point: [number, number]): string => {
  const roundedLat = Math.abs(point[0]).toFixed(2);
  const roundedLon = Math.abs(point[1]).toFixed(2);
  const lat = point[0] < 0 ? `${roundedLat}° S` : `${roundedLat}° N`;
  const lon = point[1] < 0 ? `${roundedLon}° W` : `${roundedLon}° E`;
```

Could be a good util to unit test too 🙂

fe/lib/components/LocationSuggest/LocationSuggestInput.tsx
```diff
@@ -153,6 +179,45 @@ const LocationSuggestInput = ({
...
+        <Button
+          icon={<IconLanguage />}
+          onClick={() => setShowMapDialog(true)}
+          tone={tone === 'critical' ? tone : undefined}
+          variant="soft"
+        >
+          Select on map
+        </Button>
+        <Dialog
+          id="map-dialog"
```
72636c, line 192: I think we should namespace this at minimum to avoid conflicts in the DOM:

```suggestion
          id="seek-location-suggest-map-dialog"
```

Though long term we should move towards [useId](https://reactjs.org/docs/hooks-reference.html#useid)

fe/lib/components/LocationSelectMap/styles.css.ts
```diff
@@ -0,0 +1,15 @@
...
+  transition: ['background-color', 'filter']
+    .map((prop) => `${prop} 0.2s ease-out`)
+    .join(', '),
+  ':hover': {
+    backgroundColor: vars.backgroundColor.formAccentSoft,
+  },
+  ':active': {
+    backgroundColor: vars.backgroundColor.formAccentSoft,
+    filter: 'brightness(95%)',
+  },
```
72636c, line 14: Can/should we use `formAccentSoftActive` and `formAccentSoftHover` here instead of rolling our own `filter`?

fe/lib/components/LocationSelectMap/queries.ts
```diff
@@ -0,0 +1,43 @@
...
+            parent {
+              ...locationAttributes
+            }
+          }
+        }
+      }
+    }
+  }
+  ${LOCATION_ATTRIBUTES}
+`;
```
72636c, line 34: Just curious, do we actually need this level of nesting? Is this because we feed the location in to another component that tries to display its full hierarchy?

.changeset/witty-colts-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+LocationAutosuggest: Add 'Select on map' feature
```
72636c, line 5: I think you can add another changeset like

```markdown
LocationSelectMap: Add new component
```

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
```diff
@@ -0,0 +1,191 @@
...
+    >
+      <Text>No locations found.</Text>
+    </Alert>
+  );
+
+  const hasNearestLocations = nearestLocations && nearestLocations.length > 0;
+  const suggestionsOffset: [number, number] = [-40, 130];
+
+  return (
+    <Box>
```
72636c, line 146: Would this be more descriptive if it were a `Stack space="none"`?

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
```diff
@@ -0,0 +1,191 @@
...
+      tone="info"
+      onClose={() => setShowSuggestions(false)}
+      closeLabel="Close"
+    >
+      <Text>No locations found.</Text>
+    </Alert>
+  );
+
+  const hasNearestLocations = nearestLocations && nearestLocations.length > 0;
+  const suggestionsOffset: [number, number] = [-40, 130];
```
72636c, line 143: Noob question, is there any significance to these magic numbers?

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
```diff
@@ -0,0 +1,191 @@
...
+    },
+  );
+
+  const nearestLocations =
+    data?.nearestLocations ?? previousData?.nearestLocations;
+
+  const isDesktopOrAbove = responsiveValue({
+    mobile: false,
+    desktop: true,
+  });
```
72636c, line 84: While this isn't my first preference because it imposes runtime overhead, I think the layout you're going for would be a bit difficult to accomplish in pure CSS 👍

72636c (approved):
I have a couple nits with the tablet layout, but otherwise 🚀

195741099-8de5ffdf-fb38-4de5-bd9c-ee950c394462

fe/lib/components/LocationSelectMap/index.ts
```diff
@@ -0,0 +1 @@
+export { LocationSelectMap } from './LocationSelectMap';
```
72636c, line 1: I think you'll want this export in the root `index.ts`: https://github.com/seek-oss/wingman/blob/master/fe/lib/index.ts

fe/lib/components/LocationSelectMap/queries.ts
tbolg: Yeah this level of nesting won't be used within the `LocationSelectMap` component, but is necessary for when we feed the location to other components. For example, in dev dashboard we display the full hierarchy in a `Breadcrumbs` component once a location is selected from the `LocationSuggest`

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
tbolg: These control where the suggestions will be overlayed in relation to a point. An offset of `[0,0]`, i.e, no offset, would position the suggestions with the marker like this:
Screen Shot 2022-10-17 at 5 08 44 pm

Specifying `[-40, 130] `moves the suggestions 40 pixels to the right of the marker and 130 pixels up from the marker.

tbolg: I'll add a comment on the offset so this logic is clear

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
tbolg: Yep I think it is 👍

fe/lib/components/LocationSelectMap/styles.css.ts
tbolg: Yep!

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
72636c: All good. What I meant to ask is how we decided on `-40` and `130` specifically, and whether they can be related to other numbers in the codebase (e.g. the height of the marker icon) or maybe even Braid's [spacing scale](https://seek-oss.github.io/braid-design-system/foundations/layout#spacing).

fe/lib/utils/formatPoint.ts
tbolg: Oops yep you're definitely right with that, will use this suggested solution, thanks! 😄

fe/lib/components/LocationSelectMap/LocationSelectMap.tsx
tbolg: Ah ok gotcha. Those unfortunately don't relate to anything else and I just picked them by sight. Will try see if I can get them to relate to either of those things 👍

tbolg: I've gone with Braid's [useSpace](https://seek-oss.github.io/braid-design-system/components/useSpace) hook for the X offset, but after exploring options for getting a dynamic/relative Y offset I think the only option would be doing funky stuff with refs, and that feels a bit clunky. So I'm thinking I will just keep the `130` for now

PR #866, opened 2022-12-05
createBrowserTokenMiddleware: Propagate `hirerId` request body property

72636c:


be/src/browserToken/middleware.test.ts
```diff
@@ -51,17 +51,78 @@ describe('createBrowserTokenMiddleware', () => {
...
+          {
+            "authorization": "in",
+            "hirerId": undefined,
+          },
+        ],
+      ]
+    `);
+  });
+
+  it('issues a browser token for a request with an invalid hirer ID', async () => {
```
griffithsh, line 78: Why do we want this behaviour? 🤔 I thought "scoped to a hirer" was a core feature of browser tokens.

AkiraJ48 (approved):

be/src/browserToken/middleware.test.ts
72636c: Browser tokens are always scoped to a hirer, but this test case is for a scenario where the client doesn't specify which hirer account they want to use, which could make sense if there is a 1:1 mapping between the two or the decision is made on the server side.

That said I can see how this is confusing and will remove this fallback behaviour until there's a legitimate need for it.

PR #869, opened 2022-12-06
fe: Clear local state on `hirerId` changes

72636c:


fe/lib/components/BrandSelect/BrandSelect.tsx
```diff
@@ -66,6 +72,26 @@ export const BrandSelect = ({
...
+  useEffect(() => {
+    if (
+      hasPreselected &&
+      selectedBrandId &&
+      previousVariables?.hirerId !== hirerId
+    ) {
+      // Clear stale state of the previous hirer when `hirerId` changes.
+      onSelect?.();
+
+      setSelectedBrandId(initialBrandId);
```
72636c, line 84: I'm also wondering whether we should validate that this `initialBrandId` belongs to the specified hirer 🤔

72636c:
I think this is superseded by #1016

PR #923, opened 2023-04-26
fe: Fix compilation error

72636c:
Resolves a regression from #917 due to formatting differences in GraphQL directive descriptions. `@graphql-tools/utils` does not round-trip block formatting.

https://github.com/ardatan/graphql-tools/blob/%40graphql-tools/utils%409.2.1/packages/utils/src/print-schema-with-directives.ts#L205-L211


tadhglewis (approved):
looks fine to me

PR #932, opened 2023-05-05
Call onClear in LocationSuggest

tbolg:
This was getting passed as a prop but never called when the autosuggest was cleared

### DIFF
--- a/.changeset/quick-onions-remember.md
+++ b/.changeset/quick-onions-remember.md
@@ -2,4 +2,4 @@
 'wingman-fe': patch
 ---
 
-Call onClear in LocationSuggest
+LocationSuggest: Call `onClear`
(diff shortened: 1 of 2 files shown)


fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -174,6 +175,9 @@ export const LocationSuggest = forwardRef<
             onClear={() => {
               setSelectedLocation(undefined);
               setPlaceholder('');
+              if (onClear) {
+                onClear();
+              }
```
72636c, line 180:
```suggestion
              onClear?.();
```

.changeset/quick-onions-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': patch
+---
+
+Call onClear in LocationSuggest
```
72636c, line 5:
```suggestion
LocationSuggest: Call `onClear`
```

72636c (approved):

PR #935, opened 2023-05-09
Add SalaryDetailsV2 for APAC

72636c:
This is currently implemented as a separate component version in case we want to merge this before we can release the documentation on the Developer Site, as APAC salary features are still undergoing UAT.

This will replace the component rendered at the top of this page: https://developer.seek.com/use-cases/job-posting/salary


fe/lib/components/SalaryDetails/v2/README.md
```diff
@@ -0,0 +1,98 @@
...
+    basisCode: '',
+    currency: 'AUD',
+    description: '',
+    intervalCode: '',
+    maximumAmount: null,
+    minimumAmount: '',
+    payType: '',
+  });
+
+  const onBlur = (item) => {
```
tbolg, line 51: Thinking about this discriminated union logic again, I'm thinking we can keep this as is. I like the way this provides an easy way to act on specific changes to certain data in the form and is flexible to whether you want to store the `remuneration` state as one object or in individual states. If we package the `onBlur`'s return into a single object, the example would be a bit cleaner but then we lose out on that flexibility.

fe/lib/components/SalaryDetails/v2/SalaryDetails.tsx
```diff
@@ -0,0 +1,230 @@
...
+      | SalaryDescriptionChange,
+  ) => void;
+}
+
+export const SalaryDetails = (props: SalaryDetailsProps) => {
+  const id = useId();
+
+  const {
+    errors,
+    initialCurrency,
```
tbolg, line 57: Should we add a `useEffect` in for these initial values? I think this could be useful, e.g., if someone changes the location of their job and we want to set a new initial currency

72636c: I've given this a shot in [49bd098](https://github.com/seek-oss/wingman/pull/935/commits/49bd098a13ad66cc8c1ee231d309afbf03a752d6); the way our internal codebase currently integrates this component may make things challenging but let's see how it works out 🙂

72636c:
Actually, I'll make this a breaking change

72636c:
Screenshot 2023-05-24 at 2 26 41 pm

AkiraJ48 (approved):
🚀

PR #947, opened 2023-06-19
Use sku 12 and Storybook 7

tbolg:


tadhglewis (commented):
Should we upgrade everything at once? https://github.com/seek-oss/scoobie/pull/521#pullrequestreview-1487034769

fe/lib/components/LocationLookup/LocationLookup.stories.tsx
```diff
@@ -1,36 +1,20 @@
...
-  defaultArgs,
-} from '../../storybook/controls';
-import { BraidStorybookProvider } from '../../storybook/decorators';
-
-import { LocationLookup as Component } from './LocationLookup';
-import { MockLocationLookup } from './LocationLookup.mock';
+import { MockLocationLookup as Component } from './LocationLookup.mock';
 
 export default {
+  title: 'Job Posting/Locations/LocationLookup',
```
72636c, line 8: Do we know if this migration affects the Storybook URLs?

If so we definitely need a release and have to update the manual fragments in the `.mock.tsx` components:

https://github.com/seek-oss/wingman/blob/9332027d88f56be48d1fd3288bb96db0d1ec23a0/fe/lib/components/LocationLookup/LocationLookup.mock.tsx#L32

fe/lib/components/SalaryDetails/SalaryDetails.stories.tsx
```diff
@@ -52,16 +44,8 @@ export default {
       control: { type: 'text' },
     },
   },
-  component: SalaryDetailsComponent,
-  title: 'Job Posting/Salary details',
```
72636c, line 56: This one looks to be a definite change

fe/.storybook/preview.tsx
```diff
@@ -0,0 +1,58 @@
...
+      },
+    },
+  },
+  decorators: [
+    (Story, { globals }) => (
+      <BraidProvider theme={THEMES[globals.theme as BraidThemeName]}>
+        <Box paddingX="gutter" paddingY="large">
+          <Story />
+        </Box>
+      </BraidProvider>
```
72636c, line 55: Are we intentionally removing `ContentBlock`?

fe/.storybook/preview.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export default {
+  globalTypes: {
+    theme: {
+      description: 'Braid theme to use',
+      defaultValue: 'seekJobs',
+      toolbar: {
+        title: 'Theme',
+        icon: 'contrast',
+        items: ['seekJobs', 'apac', 'docs', 'wireframe'],
```
72636c, line 45: Can we remove `apac`?

.changeset/itchy-sloths-return.md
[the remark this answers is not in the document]
tbolg: I don't think they will end up in the final bundle. I guess the only point of the change set is so that Sku 12 gets pulled through for package consumers. I can remove Storybook 7 from this messaging 👍

fe/lib/components/LocationLookup/LocationLookup.stories.tsx
tbolg: The URLs should be unaffected by this migration 👍

fe/lib/components/SalaryDetails/SalaryDetails.stories.tsx
tbolg: Ah yeah that and the `SeekApiResponse` story will now have different titles. I changed the title here so the nesting would line up with everything else, but I'll change it back to how it originally was to fix this. The changes the `SeekApiResponse shouldn't matter I don't think since theres no mock component for it.

tbolg: Nope, I accidentally removed it, will add back in 👍

fe/.storybook/preview.tsx
tbolg: Nope, this was an accident, will fix it up 👍

.changeset/itchy-sloths-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+Use sku 12
```
72636c, line 5: This is also technically inconsequential because it's a dev dependency, but I assume we should do a release anyhow just to surface any issues

72636c (approved):
🙇

PR #959, opened 2023-07-05
Update `SalaryDetails` to query currencies and pay types

AkiraJ48:
This is in reference to https://github.com/SEEK-Jobs/indie-ryanair-frontend/pull/883#pullrequestreview-1513491646 and implements the queries in https://github.com/SEEK-Jobs/indie-graphql-server/pull/2249.

### DIFF
--- a/fe/lib/components/SalaryDetails/SalaryDetails.tsx
+++ b/fe/lib/components/SalaryDetails/SalaryDetails.tsx
@@ -23,14 +23,12 @@ import type {
 } from '../../types/seekApi.graphql';
 
 import { CURRENCIES, PAY_TYPES } from './queries';
-import {
-  type PayRangeChange,
-  type PayTypeChange,
-  SALARY_CURRENCIES,
-  type SalaryCurrency,
-  type SalaryCurrencyChange,
-  type SalaryDescriptionChange,
-  type SalaryError,
+import type {
+  PayRangeChange,
+  PayTypeChange,
+  SalaryCurrencyChange,
+  SalaryDescriptionChange,
+  SalaryError,
 } from './types';
 import {
   validateDescription,
@@ -44,8 +42,9 @@ export const MAX_CHAR_LIMIT = 50;
 
 export interface SalaryDetailsProps {
   client: ApolloClient<unknown>;
+  currencyUsageTypeCode?: string;
   errors?: SalaryError;
-  initialCurrency?: SalaryCurrency;
+  initialCurrency?: string;
   initialDescription?: string;
   initialMaximumAmount?: string;
   initialMinimumAmount?: string;
@@ -75,8 +74,9 @@ function useEffectfulState<T>(initialState: T) {
 
 export const SalaryDetails = ({
   client,
+  currencyUsageTypeCode = 'SEEKMarket',
   errors,
-  initialCurrency = SALARY_CURRENCIES.default,
+  initialCurrency = 'AUD',
   initialDescription = '',
   initialMaximumAmount = '',
   initialMinimumAmount = '',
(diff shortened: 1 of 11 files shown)


fe/lib/components/SalaryDetails/SalaryDetails.tsx
```diff
@@ -87,6 +122,23 @@ export const SalaryDetails = ({
     min: validateMinAmount(min, errors),
   };
 
+  if (
+    payTypesLoading ||
+    currenciesLoading ||
+    payTypesData === undefined ||
+    currenciesData === undefined
+  ) {
+    return <Loader />;
```
72636c, line 131: Nit: maybe not for this PR, but do we want a nicer loading state that results in a smaller layout shift when the data is finally loaded?

I think "Pay details" as a heading for example would be useful to keep around even while loading.

fe/lib/components/SalaryDetails/SalaryDetails.tsx
```diff
@@ -87,6 +122,23 @@ export const SalaryDetails = ({
     min: validateMinAmount(min, errors),
   };
 
+  if (
+    payTypesLoading ||
+    currenciesLoading ||
+    payTypesData === undefined ||
+    currenciesData === undefined
```
72636c, line 129: Do these need to explicitly test for `undefined`, or can we treat `null`s the same way?

fe/lib/components/SalaryDetails/SalaryDetails.tsx
```diff
@@ -87,6 +122,23 @@ export const SalaryDetails = ({
...
+    return <Loader />;
+  }
+
+  const payTypes = payTypesData.payTypes.reduce<Record<string, PayType>>(
+    (acc, data) => ({
+      ...acc,
+      [`${data.basisCode}.${data.intervalCode}`]: data,
+    }),
+    {},
+  );
```
72636c, line 140: Is this pass necessary? I would expect the query response to be deduplicated and in a reasonable order.

72636c (approved):
🙇

fe/lib/components/SalaryDetails/SalaryDetails.tsx
AkiraJ48: Do you mean to just use the original response? I mainly adjust the shape so that it was easier for https://github.com/seek-oss/wingman/pull/959/files/448d70dbd088ab8055c79f4e187aac0d3580b146#diff-de34c4e404c87e05543daa353d41e70903945fea108a2430232043cd03912cabR266

fe/lib/components/SalaryDetails/SalaryDetails.tsx
AkiraJ48: Realistically we should never get a `null` response with these 2 queries, if we did something went very wrong 😅 .

fe/lib/components/SalaryDetails/SalaryDetails.tsx
72636c: Oh I see, I was only looking at the `Object.entries` usage below. Would it make sense to change this to a `payType = data.payTypes.find((p) => p.label === payTypeLabel)` and simplify that part to `data.payTypes.map(...)`?

72636c (approved):

fe/lib/components/SalaryDetails/SalaryDetails.tsx
72636c: Is the specific behaviour around `undefined` and not `null` guaranteed by the GraphQL server/client? If not I'd just err on the side of treating them the same in case someone decides to change the implementation in future.

fe/lib/components/SalaryDetails/SalaryDetails.tsx
AkiraJ48: Did you mean something like https://github.com/seek-oss/wingman/pull/959/commits/564ebf85389d8d823aa392403f875353a246079b?

72636c: Ooh I see, there's a bit of complexity as we need the codes and the labels in different places. I think I'm fine with either approach having seen that, thanks for looking into this 🙇

72636c (approved):

PR #1009, opened 2023-09-22
Refactor useEffectfulState to cater reset input UI rerendering

jeffchang1117:
Right now, the minimum and maximum values are managed in the local useState input that is initialized from the props. 
This change is to ensure it got rerendered when the max & min values from the props values were reset

### DIFF
--- a/.changeset/slimy-guests-melt.md
+++ b/.changeset/slimy-guests-melt.md
@@ -1,5 +1,5 @@
 ---
-'wingman-fe': minor
+'wingman-fe': patch
 ---
 
-feat: Refactor useEffectfulState hook to accept empty string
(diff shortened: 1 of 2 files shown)


.changeset/slimy-guests-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'wingman-fe': minor
+---
+
+feat: Refactor useEffectfulState hook to accept empty string
```
72636c, line 5: I think this can be considered a bug fix:

```suggestion
'wingman-fe': patch
---

SalaryDetails: Update state fields on empty strings
```

72636c (approved):

PR #1016, opened 2023-10-09
feat(LocationSuggest): Reset location on hirer change

tbolg:
Adds logic to `LocationSuggest` so that it clears when the `hirerId` changes. This is because the locations returned by the location suggestions query vary depending on the hirer.

### DIFF
--- a/fe/lib/components/LocationSuggest/LocationSuggest.tsx
+++ b/fe/lib/components/LocationSuggest/LocationSuggest.tsx
@@ -115,7 +115,7 @@ export const LocationSuggest = forwardRef<
       setSelectedLocation(undefined);
       setPlaceholder('');
       onClear?.();
-    }, [hirerId, onClear])
+    }, [hirerId, onClear]);
 
     const { data: initialLocation } = useQuery<
       LocationQuery,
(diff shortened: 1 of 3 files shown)


fe/lib/components/LocationSuggest/LocationSuggest.tsx
```diff
@@ -107,6 +107,16 @@ export const LocationSuggest = forwardRef<
...
 
+    useEffect(() => {
+      /**
+       * Location suggestions vary depending on the hirer.
+       * Hence, we should clear the selected location when the hirer changes.
+       */
+      setSelectedLocation(undefined);
+      setPlaceholder('');
+      onClear?.();
+    }, [hirerId, onClear]);
```
72636c, line 118: Does this reset the hirer ID whenever `onClear` changes? Is that desirable?

tbolg: I'm guessing you meant resets the location? Fixed this up so that it will only change if the `hirerId` changes

72636c (approved):
🙇

fe/lib/components/LocationSuggest/LocationSuggest.tsx
72636c: 🤦 Yes, thanks!

PR #1051, opened 2024-01-01
Upgrade to apollo v4

AaronMoat:


be/src/seekGraph/middleware.ts
```diff
@@ -24,26 +24,34 @@ export const createSeekGraphMiddleware = async ({
...
     ],
     schema,
   });
 
   await server.start();
 
-  return server.getMiddleware({ path });
+  const router = new Router().post(
+    path,
+    bodyParser(),
```
72636c, line 50: Should we set `enableTypes: ['json']`:

https://www.npmjs.com/package/koa-bodyparser#options

72636c (approved):

PR #1058, opened 2024-01-16
Update @apollo/client version to 3.8.9

AhmedMagedElhawary:


AaronMoat:
Discussed offline we can actually solve this with lock file regeneration

PR #1104, opened 2024-04-23
fix(LocationSuggestInput): Use IconGlobe

tbolg:
`Iconlanguage` used to be a globe but has recently changed. This makes it's use in `LocationSuggestInput` look odd now (see screenshot). Hence, I have replaced with `IconGlobe`. 

Screenshot 2024-04-23 at 1 43 56 PM


72636c (approved):

AaronMoat (approved):

72636c:
Reminder that this repo uses Changesets, and you may need to bump the minimum Braid version

tadhglewis (approved):

.changeset/two-paws-double.md
```diff
@@ -0,0 +1,6 @@
+---
+'wingman-fe': minor
```
AaronMoat, line 2:
```suggestion
'wingman-fe': major
```

I know it's silly, but it is accurate if we're bumping a peer dep requirement

tbolg: Yeah you're right, I always mess this up 😅

AaronMoat (approved):

PR #1105, opened 2024-04-23
Remove backend code

AaronMoat:
SEEK internally only has one consumer of wingman-be, and find certain things hard to iterate on with this separation. We also don't believe parts are representative of what a model ATS backend would look like, and may lead to confusion. Given we strongly advise not depending on the wingman-be package, there should be no impact to partners here. 

This could be followed up with a de-monorepoing if we in the long term find no use in having any subpackages


72636c (approved):
🔥

jeffchang1117 (approved):

tadhglewis (approved):

PR #1194, opened 2024-10-25
Mainify

samchungy:
https://github.com/seek-oss/wingman/blob/d80086b606bd4f1559a9048504276ec04cf5a154/.github/workflows/release.yml#L6-L7


AaronMoat (approved):
👍

PR #1251, opened 2025-02-13
Wrap `InlineCode` with `Text`

72636c:
`Text` sets the right color mode and stops the element from stretching to full width.


72636c:
🤠

PR #1343, opened 2025-10-07
Revert "fix: @types/uuid"

samchungy:
Reverts seek-oss/wingman#1333

Crypto is a Node.js library, we can use window.crypto.randomUUID() down the line but our browser support policy doesn't allow that yet: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility

### DIFF
--- a/fe/package.json
+++ b/fe/package.json
@@ -6,12 +6,14 @@
     "@apollo/client": "^4.0.0",
     "@graphql-tools/schema": "^10.0.0",
     "@types/autosuggest-highlight": "^3.2.3",
+    "@types/uuid": "^10.0.0",
     "autosuggest-highlight": "^3.2.0",
     "pigeon-maps": "^0.22.0",
     "react-fast-compare": "^3.2.0",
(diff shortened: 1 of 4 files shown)


fe/package.json
```diff
@@ -6,12 +6,14 @@
     "@apollo/client": "^4.0.0",
     "@graphql-tools/schema": "^10.0.0",
     "@types/autosuggest-highlight": "^3.2.3",
+    "@types/uuid": "^10.0.0",
```
zbrydon, line 9:
```suggestion
```

72636c:
Which browser?

[This](https://developer.seek.com/introduction/browser-support) is our policy, and the MDN page lists versions that were all released by 2022.

samchungy:
> Which browser?
> 
> [This](https://developer.seek.com/introduction/browser-support) is our policy, and the MDN page lists versions that were all released by 2022.

Oh didn't realise we maintain a different list to SEEK?

